PR #12, opened 2019-06-16
Add docs using Sphinx

OrangeTux:
A compiled version of the docs can be found here:https://ocpp.readthedocs.io/en/latest/


nshakhat:
Yeees!!! We are going to use Sphinx in ORC as well

README.rst
```diff
@@ -0,0 +1,144 @@
...
+   from datetime import datetime
+
+   from ocpp.routing import on
+   from ocpp.v16 import ChargePoint as cp
+   from ocpp.v16.enums import Action, RegistrationStatus
+   from ocpp.v16 import call_result
+
+
+   class ChargePoint(cp):
+       @on(Action.BootNotification)
```
tropxy: Cool! I see that later you explain in more depth the on decoratir. However, I didnt see any mention to the after one....shouldnt we introduce the after decorator as well?

OrangeTux: I don't think the quick start section is the README.md is the right place for explaining code. But maybe I should add a link to the corresponding pages of the docs.

tropxy: But I was not requesting for further explanation. I was requesting for a mention if not here somewhere, to the @after decorator as well

OrangeTux: Fair point. I didn't read it properly. I've updated the PR and addresses your comment.

pyproject.toml
```diff
@@ -6,7 +6,7 @@ authors = [
 	"Auke Willem Oosterhoff <aukewillem.oosterhoff@mobilityhouse.com>"
```
tropxy: Not being pretentious, but would be nice to mention my name as well

OrangeTux: Sure, of course! I was thinking to add all addresses that where in the AUTHORS file in the closed-source version of this repo. What do you think?

tropxy: Yeah, sounds good ;)

pyproject.toml
```diff
@@ -3,10 +3,19 @@ name = "ocpp"
 version = "0.2.0"
 description = "Python package implementing the JSON version of the Open Charge Point Protocol (OCPP)."
 authors = [
-	"Auke Willem Oosterhoff <aukewillem.oosterhoff@mobilityhouse.com>"
+	"Andre Duarte <andre.duarte@mobilityhouse.com>",
+	"Auke Willem Oosterhoff <aukewillem.oosterhoff@mobilityhouse.com>",
+	"Greg Lutostanski <greg.lutostanski@mobilityhouse.com>",
+	"Jonathan Herrmann <jonathan.herrmann@mobilityhouse.com>",
+	"Laysa Uchoa <laysa.uchoadasilva@mobilityhouse.com>",
+	"Oz N Tiram <oz.tiram@mobilityhouse.com>",
```
tropxy: Oz does not work for Mobility house anymore...does it make sense to leave it here?

OrangeTux: I think it is fair to list everyone who worked on it. We could remove the email address though. That is likely not going to be an active address.

tropxy: yup, agree!

proelke: Maybe its better to link github accounts or his github at least?

tropxy: yup, mabe that it the most stable option...what do you think @OrangeTux ?

PR #27, opened 2019-09-24
Use create_task when executing the 'after' handler coroutine

proelke:
If a call is used in the 'after' handler, it can block the event loop, so it is better to run the coroutine in a task.

### DIFF
--- a/ocpp/v16/charge_point.py
+++ b/ocpp/v16/charge_point.py
@@ -207,7 +207,7 @@ class ChargePoint:
             handler = handlers['_after_action']
             # Create task to avoid blocking when making a call inside the
             # after handler
-            asyncio.create_task(
+            asyncio.ensure_future(
                 asyncio.coroutine(handler)(**snake_case_payload))
         except KeyError:
             # '_on_after' hooks are not required. Therefore ignore exception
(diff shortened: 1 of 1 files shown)


ocpp/v16/charge_point.py
```diff
@@ -205,7 +205,10 @@ def __init__(self, id, connection, response_timeout=30):
 
         try:
             handler = handlers['_after_action']
-            await asyncio.coroutine(handler)(**snake_case_payload)
+            # Create task to avoid blocking when making a call inside the
+            # after handler
+            asyncio.create_task(
```
tropxy: This fails in python3.6. instead you can use asyncio.ensure_future(coro()), which works in all python versions

doumelas: check  the doc of asyncio :
[https://docs.python.org/3/library/asyncio-task.html#id4](url)

proelke: Ughhh.

proelke:
Updated it to use ensure_future. @tropxy

tropxy:
Did you test this? Just to be sure :P

proelke:
I did in-fact test it with @OrangeTux.

PR #30, opened 2019-10-18
Add message classes for OCPP 2.0

OrangeTux:
@tropxy I am sorry to invalidate your work done on implementing the message classes for OCPP 2.0.
While going over your code I saw that sill several classes where missing. Instead of manually figuring out which message was missing I wrote a small script that generates all classes using the JSON schema's. That script is included  in this PR, but you don't need to review.


ocpp/v20/schemas/SetMonitoringLevelRequest_v1p0.json
```diff
@@ -0,0 +1,15 @@
...
+  "additionalProperties": true,
+  "properties": {
+    "severity": {
+      "type": "integer"
+    }
+  },
+  "required": [
+    "severity"
+  ]
+}
```
tropxy: misses a new line

ocpp/v20/schemas/SetMonitoringLevelResponse_v1p0.json
```diff
@@ -0,0 +1,20 @@
...
+      "enum": [
+        "Accepted",
+        "Rejected"
+      ]
+    }
+  },
+  "required": [
+    "status"
+  ]
+}
```
tropxy: here too is missing

ocpp/v20/call.py
```diff
@@ -0,0 +1,421 @@
...
+
+@dataclass
+class FirmwareStatusNotificationPayload:
+    status: str
+    request_id: int
+
+
+@dataclass
+class Get15118EVCertificatePayload:
+    _15118_schema_version: str
```
tropxy: This one I wanted to discuss with you. I added an underscore before the number, however I dont know if when we convert from camel case to snake case if it will convert it with the underscore or not

scripts/schema_to_dataclass.py
```diff
@@ -0,0 +1,166 @@
...
+    "array": "List",
+    "integer": "int",
+    "string": "str",
+    "number": "int",
+    "boolean": "bool",
+    "any": "Any",
+}
+
+
+def create_dataclass(name):
```
tropxy: it is worth it to wrap the dataclass into the create_dataclass when we could just use directly dataclass()?

scripts/schema_to_dataclass.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def __str__(self):
+        name = self.name
+        if not re.match("^[a-zA-Z_]", self.name):
+            name = "_" + self.name
+
+        definition = f"    {name}: {self.type}"
+        if self.required is True:
+            # if self.type == "Dict":
+                # definition += " = field(default_factory=dict)\n"
```
tropxy: why did you comment it out?

ocpp/v20/schemas/PublishFirmwareStatusNotificationRequest_v1p0.json
```diff
@@ -0,0 +1,36 @@
...
+        "DownloadScheduled",
+        "DownloadPaused",
+        "PublishFailed",
+        "Published",
+        "InvalidChecksum",
+        "ChecksumVerified"
+      ]
+    },
+    "location": {
+      "type": "array",
```
tropxy: according to the spec section 1.43.2 page 315, the PublishFirmwareStatusNotification attribute "location" has a string type. Therefore, this schema is wrong and hast to be changed to type: string with a maxlength of 512

ocpp/v20/call.py
OrangeTux: We discussed offline that we don't like it, but also have not better solution right now.

scripts/schema_to_dataclass.py
OrangeTux: I think the script can be improved in several ways. This PR isn't about the script, more about the classes it generates. 

I added the script in order not to lose it. For this PR I ignore comments on the script if you don't mind.

scripts/schema_to_dataclass.py
OrangeTux: I think the script can be improved in several ways. This PR isn't about the script, more about the classes it generates. 

I added the script in order not to lose it. For this PR I ignore comments on the script if you don't mind.

ocpp/v20/schemas/PublishFirmwareStatusNotificationRequest_v1p0.json
OrangeTux: Good catch!

scripts/schema_to_dataclass.py
tropxy: okidoki ;)

PR #33, opened 2019-10-28
Create generic chargepoint

OrangeTux:


ocpp/charge_point.py
```diff
@@ -257,15 +263,17 @@ def __init__(self, id, connection, response_timeout=30):
         if response.message_type_id == MessageType.CallError:
             LOGGER.warning("Received a CALLError: %s'", response)
             return
+        else:
+            response.action = call.action
+            validate_payload(response, self.ocpp_version)
```
tropxy: where is the self.ocpp_version set?

ocpp/messages.py
```diff
@@ -50,50 +61,71 @@ def pack(msg):
...
-def get_schema(name):
+def get_schema(message_type_id, action, ocpp_version):
     """
     Read schema from disk and return in. Reads will be cached for performance
     reasons.
+
     """
-    if name in _schemas:
-        return _schemas[name]
+    if ocpp_version == "1.6":
```
tropxy: we could do some string operations here and substitute the if elses clauses by:

```python
schemas_dir = "v" + ocpp_version.replace(".", "")
```
so then if new versions come we dont need to do a train of if clauses
what do you think?

tropxy: for the error raise, maybe we can just check if the ocpp_version is in the list ["1.6", "2.0", ...]

ocpp/messages.py
```diff
@@ -50,50 +61,71 @@ def pack(msg):
...
+        schema = get_schema(
+            message.message_type_id, message.action, ocpp_version
+        )
     except (OSError, json.JSONDecodeError) as e:
         raise ValidationError("Failed to load validation schema for action "
-                              f"'{action}': {e}")
+                              f"'{message.action}': {e}")
 
-    if action in ['SetChargingProfile', 'RemoteStartTransaction']:
+    if message.action in ['SetChargingProfile', 'RemoteStartTransaction']:
```
tropxy: for ocpp version 1.6 we have "RemoteStartTransaction", but for 2.0 we have the new "RequestStartTransaction". So, this option should be added

ocpp/v20/schemas/AuthorizeRequest_v1p0.json
```diff
@@ -1,6 +1,5 @@
 ﻿{
```
tropxy: Do you think there is a way we can inform why we have removed this id reference, so next time with even new schemas we know we have to remove it?

ocpp/charge_point.py
OrangeTux: In `ocpp.{v16,v20}.ChargePoint`

ocpp/messages.py
OrangeTux: Sure, I like it.

OrangeTux: Sure, I like it.

ocpp/v20/schemas/AuthorizeRequest_v1p0.json
tropxy: schade

Issue #43, opened 2019-11-12
Fix validation of payloads containing floats


pagetable:
What are the todo and special actions in validate_payload() supposed to be? Are they to deal with the 'multipleof' problem of jsonschema check?

https://github.com/mobilityhouse/ocpp/blob/75dd91fb80f34c401365061e90db9907a9faac33/ocpp/messages.py#L118-L124

Besides, I am not sure if it is proper to ask this question here, if not, please feel free to delete this issue.

Thank you.

OrangeTux:
Hello @pagetable 

Nice catch! Here is a comment I found in our internal code base related to this issue:

``` python
# These 3 schedules have fields of type floats. The JSON schema defines
# a certain precision for these fields of 1 decimal. A value of 21.4 is
# valid, whereas a value if 4.11 is not.
#
# The problem is that Python's internal representation of 21.4 might
# have more than 1 decimal. It might be 21.399999999999995. This would
# make the validation fail, although the payload is correct. This is a
# known issue with jsonschemas, see:
# https://github.com/Julian/jsonschema/issues/247
#
# This issue can be fixed by using a different parser for floats than
# the default one that is used.
#
# Both the schema and the payload must be parsed using the different
# parser for floats.
```

I will implement it soon.

PR #45, opened 2019-11-13
Fix validation of payloads containing floats

OrangeTux:
The validation of payloads using jsonschemas could fail when the payload
contained a float. This problem is described in this issue:
https://github.com/Julian/jsonschema/issues/247

This commit implements a work around for this issue by changing the
float parser for certain payloads from `float()` to `decimal.Decimal()`.

Fixes: #43

### DIFF
--- a/ocpp/messages.py
+++ b/ocpp/messages.py
@@ -2,6 +2,7 @@
 also contain some helper functions for packing and unpacking messages.  """
 import os
 import json
+import decimal
 from dataclasses import asdict, is_dataclass
 
 from jsonschema import validate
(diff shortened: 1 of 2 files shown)


ocpp/messages.py
```diff
@@ -32,7 +33,7 @@ def unpack(msg):
     Unpacks a message into either a Call, CallError or CallResult.
     """
     try:
-        msg = json.loads(msg)
+        msg = json.loads(msg, parse_float=decimal.Decimal)
```
proelke: Is it ok to always parse floats here but only schemas in a few instances?

OrangeTux: Good catch, I need to verify that.

proelke: I started working on this a bit yesterday, and I got stuck on that part of porting the code over. If it's an issue I have an idea.

OrangeTux: I've removed the argument `parse_float=decimal.Decimal`. It is not needed in this place.

If a message requires special validation regarding the floats the message is reparsed again using `parse_float=decimal.Decimal`. See line 138 of this file.

Issue #46, opened 2019-11-15
self._call_lock never released when trying to send payload on a closed socket


psauer:
https://github.com/mobilityhouse/ocpp/blob/75dd91fb80f34c401365061e90db9907a9faac33/ocpp/charge_point.py#L251

If you try and send a payload, but the socket connection is terminated by the server just before trying to send, a `websockets.exceptions.ConnectionClosedError` is raised but never caught here. This leads to `self._call_lock.release()` never being called, making it impossible to send data again.

OrangeTux:
Good catch! Thanks for reporting it

PR #47, opened 2019-11-15
Fix potential deadlock

OrangeTux:
ocpp.ChargePoint.call() implements flow control as defined in the OCPP
specification. It makes sure that no call can be send after an answer on
the previous request has been received.

A lock is used to implement flow control. If the call `await
self._send(call.json())` crashed it the lock wouldn't be released
which could lead to a deadlock.

Thanks @psauer for reporting.

Fixes: #46

### DIFF
--- a/ocpp/charge_point.py
+++ b/ocpp/charge_point.py
@@ -248,17 +248,15 @@ class ChargePoint:
         # Use a lock to prevent make sure that only 1 message can be send at a
         # a time.
         await self._call_lock.acquire()
-        await self._send(call.to_json())
 
         try:
+            await self._send(call.to_json())
             response = \
                 await self._get_specific_response(call.unique_id,
                                                   self._response_timeout)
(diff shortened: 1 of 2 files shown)


ocpp/charge_point.py
```diff
@@ -248,18 +248,16 @@ def __init__(self, id, connection, response_timeout=30):
         # Use a lock to prevent make sure that only 1 message can be send at a
         # a time.
         await self._call_lock.acquire()
```
psauer: If you used `async with self._call_lock:`, you wouldn't need to worry about releasing the lock. In my opinion, it's a bit cleaner as well.

OrangeTux: Thanks for the suggestion! I used it in #51

OrangeTux: Thanks for the suggestion. I used it in #51

Issue #54, opened 2019-12-02
Add option to disable validation against JSON schemas


OrangeTux:
TMH has integrated a charge point which isn't fully compliant with the OCPP 1.6 specification. The MeterValues requests send by the charge point contains an 'illegal' measurand. Receiving a MeterValues request will fail with:

``` python
Traceback (most recent call last):
File "/venv/lib/python3.7/site-packages/ocpp/messages.py", line 127, in validate_payload
     validate(message.payload, schema)
    File "/venv/lib/python3.7/site-packages/jsonschema/validators.py", line 899, in validate
      raise error
  jsonschema.exceptions.ValidationError: 'Request.Energy' is not one of ['Energy.Active.Export.Register', 'Energy.Active.Import.Register', 'Energy.Reactive.Export.Register', 'Energy.Reactive.Import.Register', 'Energy.Active.Export.Interval', 'Energy.Active.Import.Interval', 'Energy.Reactive.Export.Interval', 'Energy.Reactive.Import.Interval', 'Power.Active.Export', 'Power.Active.Import', 'Power.Offered', 'Power.Reactive.Export', 'Power.Reactive.Import', 'Power.Factor', 'Current.Import', 'Current.Export', 'Current.Offered', 'Voltage', 'Frequency', 'Temperature', 'SoC', 'RPM']
```

In order to support chargers like this it will be possible to disable JSON schema validation of incoming requests. The validation can be disabled by passing `skip_schema_validation=True` to a `@on` decorator. This will disable validation for that specific request and the response.

``` python
@on(MeterValues, skip_schema_validation=True)
def on_meter_values(self, *args, **kwargs):
    pass
```

PR #56, opened 2019-12-03
Add arg `skip_schema_validation` to @on()

OrangeTux:
The argument can be used to skip validation of a request and response.
It defaults to `False`, so by default validation is enabled.

The reasoning for this change is that TMH has charge points which use
measurands inside MeterValues messages which are not compliant with the
OCPP 1.6 specification. Therefore the validation of these messages fail.

With this new argument validation of certain messages can be disabled.

Fixes: #54

### DIFF
--- a/ocpp/charge_point.py
+++ b/ocpp/charge_point.py
@@ -140,11 +140,6 @@ class ChargePoint:
             return
 
         if msg.message_type_id == MessageType.Call:
-            # Call's can be validated right away because the 'action' is known.
-            # The 'action' is required to get the correct schema.
-            #
-            # CallResult's don't have an action field. The action must be
-            # deducted from corresponding Call.
             await self._handle_call(msg)
         elif msg.message_type_id in \
                 [MessageType.CallResult, MessageType.CallError]:
--- a/ocpp/charge_point.py
+++ b/ocpp/charge_point.py
@@ -140,12 +140,11 @@ class ChargePoint:
             return
 
         if msg.message_type_id == MessageType.Call:
-            # Call's can be validated right away because the 'action' is know.
+            # Call's can be validated right away because the 'action' is known.
             # The 'action' is required to get the correct schema.
             #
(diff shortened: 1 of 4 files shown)


ocpp/charge_point.py
```diff
@@ -140,12 +140,11 @@ def __init__(self, id, connection, response_timeout=30):
             return
 
         if msg.message_type_id == MessageType.Call:
-            # Call's can be validated right away because the 'action' is know.
+            # Call's can be validated right away because the 'action' is known.
```
proelke: This comment seems a little out of place to me since we moved the validation step.

OrangeTux: Good catch!

ocpp/routing.py
```diff
@@ -1,7 +1,7 @@
 import functools
 
 
-def on(action):
+def on(action, *, skip_schema_validation=False):
```
proelke: Where does the asterisk come into play?

OrangeTux: All arguments after the aterisk can only used with as keyword arguments.

So the following call would raise a TypeError:

``` python
def foo(*, a):
    pass

foo(1)  # will raise a TypeError
```

Instead it is required to use keyword arguments:

``` python
foo(a=1)  # works fine
```

 Without aterisk the `@on()` decorator could be used like this:

``` python
@on("MeterValues", True)
def on_meter_values(*args, **kwargs):
   pass
```

Without the keyword it is hard to guess what the meaning is of the second argument. I added the aterisk to make it explicit to disable validation. You can either use `skip_schema_validation`s default value of `False`. Or, if you want to disable the validation, you can set it to true by keyword arguments:

``` python
@on("MeterValues", skip_schema_validation=True)
def on_meter_values(*args, **kwargs):
   pass
```

proelke: Interesting. Thanks for the explanation.

PR #58, opened 2019-12-04
Updating to the new versions OCPP1.6 schemas

### LINKED ISSUE #57: Implement errata v4 for OCPP 1.6
In the OCPP 1.6 schemas for MeterValues.req, OCA misspelled Celsius as Celcius.

In the OCPP  1.6 erratas, the recommended action to handle this is:

* For ChargingStations: send Celsius (ie the correct spelling), Celcius is also allowed.
* For CentralSystems: accept both Celsius and Celcius.

There are new schema files available from OCA.

darander:
OCA has released new versions of the OCPP1.6 schemas together
with erratas.

This is the latest version.

Fixes #57 (adding Celsius with correct spelling) and more.

Signed-off-by: Anders Darander <anders@chargestorm.se>

### DIFF
--- a/ocpp/v16/schemas/StopTransaction.json
+++ b/ocpp/v16/schemas/StopTransaction.json
@@ -17,144 +17,154 @@
         "transactionId": {
             "type": "integer"
         },
-		"reason": {
-			"type": "string",
-			"enum": [
-				"EmergencyStop",
-				"EVDisconnected",
-				"HardReset",
-				"Local",
-				"Other",
-				"PowerLoss",
-				"Reboot",
-				"Remote",
-				"SoftReset",
-				"UnlockCommand",
-				"DeAuthorized"
-			]
-		},
-		"transactionData": {
-			"type": "array",
-			"items": {
-				"type": "object",
-				"properties": {
-					"timestamp": {
-						"type": "string",
-						"format": "date-time"
-					},
-					"sampledValue": {
-						"type": "array",
-						"items": {
-							"type": "object",
-							"properties": {
-								"value": {
-									"type": "string"
-								},
-								"context": {
-									"type": "string",
-									"enum": [
-										"Interruption.Begin",
-										"Interruption.End",
-										"Sample.Clock",
-										"Sample.Periodic",
-										"Transaction.Begin",
-										"Transaction.End",
-										"Trigger",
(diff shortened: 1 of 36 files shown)


darander:
Thanks for the comment.

I've updated the commit to omit the id lines.

ocpp/v16/schemas/StopTransaction.json
```diff
@@ -1,166 +1,175 @@
...
+                                        "W",
+                                        "kW",
+                                        "VA",
+                                        "kVA",
+                                        "var",
+                                        "kvar",
+                                        "A",
+                                        "V",
+                                        "K",
+                                        "Celcius",
```
OrangeTux: It seems that `Celsius` is missing here.

darander: That's right.
I'll bring it back to OCA, as I prefer to not touch the schemas here. (I think it's better to use the upstream schemas if possible).

OrangeTux: Fair point.
I guess that the OCA fix the bug in a new V5 errate. We can implement the fix than after the errata has been released.

For now I am ready to merge the PR. Thanks for your help!

OrangeTux:
For some reason the builds on CirceCi haven't been run. I will try to fix that. If tests pass I'm happy to merge!

darander:
I've added Celsius to StopTransaction.json, OCA is notified, but I rather not wait for them anyway.

I've also updated the commit message with info on when the new schema files were released and which modifications we've done to them.

ocpp/v16/schemas/StopTransaction.json
tropxy: I also have contacted the OCA. It seems that they on purpose want the schemas to remain unchanged and that, for now, CS must adapt the enumerations and include both spellings and CP must correct the spelling.

OrangeTux:
@darander I fixed CircleCI's settings and now the test suite has ran. Unfortunately the build fails, can you fix it?

darander:
Fixed the build.

Thanks for getting the CircleCI working again!

OrangeTux:
Many thanks for your help!

darander:
Thanks for your help with my contribution!

PR #59, opened 2019-12-05
Test if this commit in a fork triggers a CI build

OrangeTux:


OrangeTux:
The test was successful. PR comings from forks now trigger CI builds

PR #60, opened 2019-12-05
Updating to the new versions OCPP1.6 schemas

### LINKED ISSUE #57: Implement errata v4 for OCPP 1.6

OrangeTux:
OCA has released new versions of the OCPP1.6 schemas together
with erratas.

This is the latest version published by OCA on 2019-12-05.
The only modifications from above are:
* Removal of id-fields (not handled by jsonschemas)
* Adding Celsius to the enum in StopTransaction.json.
  This was missed in the schema updates, but is according
  to the errata.

Fixes #57 (adding Celsius with correct spelling) and more.

Signed-off-by: Anders Darander <anders@chargestorm.se>

### DIFF
--- a/ocpp/v16/schemas/AuthorizeResponse.json
+++ b/ocpp/v16/schemas/AuthorizeResponse.json
@@ -6,16 +6,6 @@
         "idTagInfo": {
             "type": "object",
             "properties": {
-                "status": {
-                    "type": "string",
-                    "enum": [
-                        "Accepted",
-                        "Blocked",
-                        "Expired",
-                        "Invalid",
-                        "ConcurrentTx"
-                    ]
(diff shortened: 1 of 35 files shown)


OrangeTux:
@darander I fixed CircleCI's settings and now the test suite has ran. Unfortunately the build fails, can you fix it?

PR #42, opened 2019-11-07
OCPP 2.0 Enums and helper methods

tropxy:

PR #61, opened 2019-12-13
fix inheritance in enum classes

lsaavedr:
Warning, this change break other software with the assumption of "enum" classes are not python Enum. But it is necessary to apply the assumption of enum classes are python Enum in another framework like Django...

The main change is to obtain values, for example:

After:
>>> status=RegistrationStatus.accepted

before:
>>> status=RegistrationStatus.accepted.value

Regards,
Luis

### DIFF
--- a/docs/source/central_system.rst
+++ b/docs/source/central_system.rst
@@ -102,7 +102,7 @@ Remove the `on_connect()` handler from the code above and replace it by the foll
            return call_result.BootNotificationPayload(
                current_time=datetime.utcnow().isoformat(),
                interval=10,
-               status=RegistrationStatus.accepted
+               status=RegistrationStatus.accepted.value
            )
 
 
--- a/examples/v16/central_system.py
+++ b/examples/v16/central_system.py
@@ -23,7 +23,7 @@ class ChargePoint(cp):
         return call_result.BootNotificationPayload(
 	    current_time=datetime.utcnow().isoformat(),
 	    interval=10,
-	    status=RegistrationStatus.accepted
+	    status=RegistrationStatus.accepted.value
 	)
 
 
--- a/examples/v16/charge_point.py
+++ b/examples/v16/charge_point.py
@@ -25,7 +25,7 @@ class ChargePoint(cp):
 
         response = await self.call(request)
 
-        if response.status ==  RegistrationStatus.accepted:
+        if response.status ==  RegistrationStatus.accepted.value:
             print("Connected to central system.")
 
 
--- a/ocpp/routing.py
+++ b/ocpp/routing.py
@@ -96,17 +96,17 @@ def create_route_map(obj):
             try:
                 action = getattr(attr, option)
 
-                if action not in routes:
-                    routes[action] = {}
+                if action.value not in routes:
+                    routes[action.value] = {}
 
                 # Routes decorated with the `@on()` decorator can be configured
                 # to skip validation of the input and output. For more info see
                 # the docstring of `on()`.
                 if option == '_on_action':
-                    routes[action]['_skip_schema_validation'] = \
+                    routes[action.value]['_skip_schema_validation'] = \
                         getattr(attr, '_skip_schema_validation', False)
 
-                routes[action][option] = attr
+                routes[action.value][option] = attr
 
             except AttributeError:
                 continue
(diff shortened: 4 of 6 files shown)


OrangeTux:
Thanks for your PR. I've decided not to accept it for now.

Before this repo went open source the classes did subclass from Enum. But it was removed because it didn't add any value. The only thing is that it decreases readability in my opinion.

lsaavedr:
Yes, it decreases readability a little but if you want integrate your library with another framework like a Django you need make enum classes python's Enum, for example to make automatic choice you can't make this with actual enum class:

```python
from django.db import models
from ocpp.v16.enums import ChargingProfilePurposeType
...
class ChargePoint(models.Model):
...
    chargingProfilePurpose = models.CharField(
        max_length=64,
        choices=[(tag, tag.value) for tag in ChargingProfilePurposeType]
    )
```

regards!

OrangeTux:
Can't you solve your problem like this?

``` python
>>> [(getattr(ChargingProfilePurposeType, a), getattr(ChargingProfilePurposeType, a)) for a in dir(ChargingProfilePurposeType) if not a.startswith('__')]
[('ChargePointMaxProfile', 'ChargePointMaxProfile'), ('TxDefaultProfile', 'TxDefaultProfile'), ('TxProfile', 'TxProfile')]
```

OrangeTux:
Closed due inactivity. Feel free to reopen if needed.

lsaavedr:
Now I have more time to discuss it!

OrangeTux:
> Why subclassing from Enum is not needed (I need it in my django backend), using getattr is not a good practice 😭, Enum was proposal for this!

_Originally posted by @lsaavedr in https://github.com/mobilityhouse/ocpp/pull/42#issuecomment-596265650_

I copied @lsaavedr 's comment to this issue to keep discussion in 1 place. 

@lsaavedr As said, I currently don't see the added value of subclassing all classes in `enum.py` from `Enum`. Sure, in your specific case, if you use this package with Django, it make sense. But for non-Django users it makes less sense. It makes using enums less straightforward and decreases readability because every use the classes must be appended with `.value`. 

Also I've proposed an simple solution to make this package working with Django without change the API in a backward incompatible way.

Therefore I'm currently still against this PR.

adamchainz:
> But it was removed because it didn't add any value.

Enums add a bunch of value, such as potential for type checking with mypy, not comparing with equal with other enums, preventing bugs from accidentally mixing the wrong types (https://docs.python.org/3.8/library/enum.html#comparisons) and many utilities for enumerating allowed values etc. (https://docs.python.org/3.8/library/enum.html#how-are-enums-different).

OrangeTux:
Thanks for your comment. Although I personally see an added value of subclassing from `enum.Enum`, it's clear that you and others do see benefits. So I'm reconsidering whether or not the we should subclass from `enum.Enum`.

My biggest concern is that this change would change the API. It would be nice if we can come up with a solution that is backwards compatible. So that this code still works:

```python
  @on(Action.BootNotification)
    def on_boot_notitication(self, charge_point_vendor, charge_point_model, **kwargs):
        return call_result.BootNotificationPayload(
	    current_time=datetime.utcnow().isoformat(),
	    interval=10,
	    status=RegistrationStatus.accepted
	)
```

adamchainz:
Afaict the only thing stopping that from working is conversion to JSON doesn't work, which can be done with another "fixer" function similar to `remove_nones`:

```python
def convert_enums(dict_to_scan):
    return {
        k: (v.value if isinstance(v, Enum) else v)
        for k, v in dict_to_scan.items()
    }
```

OrangeTux:
I think I found a solution that solves my argument that I had against integrating with `enum.Enum`

> It makes using enums less straightforward and decreases readability because every use the classes must be appended with .value.

> Also I've proposed an simple solution to make this package working with Django without changing the API in a backward incompatible way.

Subclassing from both `str` and `enum.Enum` allows to use enums without having to use the `.value()` appendix.

``` python
from enum import Enum

class Action(str, Enum):
      """ An Action is a required part of a Call message. """
      Authorize = "Authorize"

assert Action.Authorize == "Authorize"
assert Action.Authorize.value == "Authorize"
```

So I'll reopen the ticket.

Issue #68, opened 2020-02-18
Sending a SetChargeProfile call fails with TypeError


OrangeTux:
Version `0.6.0` forces validation of `Call`s before the are send to the other end. The validation of `call.SetChargingProfilePayload` fails with a TypeError. The charging profile contains a value of type `decimal.Decimal` and this value cannot be serialized to JSON. The problem can be demonstrated like this:

``` python
>>> import decimal
>>> import json
>>> >>> json.dumps(decimal.Decimal(3))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/developer/.pyenv/versions/3.7.0/lib/python3.7/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/home/developer/.pyenv/versions/3.7.0/lib/python3.7/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/home/developer/.pyenv/versions/3.7.0/lib/python3.7/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/home/developer/.pyenv/versions/3.7.0/lib/python3.7/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type Decimal is not JSON serializable
```

tropxy:
After a short research it seems we have 2 options:

1) is to follow the suggestion in this python blog message that @OrangeTux found:
[https://bugs.python.org/issue16535#msg176158](https://bugs.python.org/issue16535#msg176158)

2) Add simplejson as dependency. simplejson supports natively Decimal serialization:
[https://simplejson.readthedocs.io/en/latest/](https://simplejson.readthedocs.io/en/latest/)

PR #69, opened 2020-02-19
Fix JSON encoding of decimal.Decimal

OrangeTux:
Version `0.6.0` forces validation of `Call`s before they are send to the
other end. The validation of `call.SetChargingProfilePayload` fails with
a TypeError. The charging profile contains a value of type
`decimal.Decimal` and this value cannot be serialized to JSON. The
problem can be demonstrated like this:

	>>> import decimal
	>>> import json
	>>> >>> json.dumps(decimal.Decimal(3))
	Traceback (most recent call last):
	  File "<stdin>", line 1, in <module>
	  File "/home/developer/.pyenv/versions/3.7.0/lib/python3.7/json/__init__.py", line 231, in dumps
	    return _default_encoder.encode(obj)
	  File "/home/developer/.pyenv/versions/3.7.0/lib/python3.7/json/encoder.py", line 199, in encode
	    chunks = self.iterencode(o, _one_shot=True)
	  File "/home/developer/.pyenv/versions/3.7.0/lib/python3.7/json/encoder.py", line 257, in iterencode
	    return _iterencode(o, 0)
	  File "/home/developer/.pyenv/versions/3.7.0/lib/python3.7/json/encoder.py", line 179, in default
	    raise TypeError(f'Object of type {o.__class__.__name__} '
	TypeError: Object of type Decimal is not JSON serializable

This bug has been fixed by creating a custom enconder for
`decimal.Decimal`.

Fixes: #68

### DIFF
--- a/ocpp/messages.py
+++ b/ocpp/messages.py
@@ -16,6 +16,36 @@ from ocpp.exceptions import (OCPPError, FormatViolationError,
 _schemas = {}
 
 
+class _DecimalEncoder(json.JSONEncoder):
+    """ Encode values of type `decimal.Decimal` using 1 decimal point.
+
+    A custom encoder is required because `json.dumps()` cannot encode a value
+    of type decimal.Decimal. This raises a TypeError:
+
+        >>> import decimal
+        >>> import json
+        >>> >>> json.dumps(decimal.Decimal(3))
+        Traceback (most recent call last):
+          File "<stdin>", line 1, in <module>
+          File "/home/developer/.pyenv/versions/3.7.0/lib/python3.7/json/__init__.py", line 231, in dumps  # noqa
+            return _default_encoder.encode(obj)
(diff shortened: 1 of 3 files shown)


ocpp/messages.py
```diff
@@ -203,7 +233,8 @@ def to_json(self):
         ],
             # By default json.dumps() adds a white space after every separator.
             # By setting the separator manually that can be avoided.
-            separators=(',', ':')
+            separators=(',', ':'),
+            cls=_DecimalEncoder
```
proelke: I'm wondering if this should just be a constant at the top to reduce some lines of code. Not really a huge deal though.

OrangeTux: What do you mean? I am afraid I don't get it.

proelke: Ah nevermind. I miss read the code. I thought that was a list you were passing in as a param, not params to the function. Disregard.

PR #88, opened 2020-05-06
Added support for OCPP 2.0.1 

tropxy:
* ~custom_data field is introduced by schemas_to_class.py script (script must be adapted)~ [SOLVED]

* ~Schemas report an issue when going through validation~: [SOLVED]
```python
self = <jsonschema.validators.RefResolver object at 0x104f84bd0>, url = ''

    def resolve_from_url(self, url):
        """
        Resolve the given remote URL.
        """
        url, fragment = urldefrag(url)
        try:
            document = self.store[url]
        except KeyError:
            try:
                document = self.resolve_remote(url)
            except Exception as exc:
>               raise exceptions.RefResolutionError(exc)
E               jsonschema.exceptions.RefResolutionError: unknown url type: ''

/usr/local/lib/python3.7/site-packages/jsonschema/validators.py:779: RefResolutionError
```
* ~Enums are not complete~ [SOLVED]

The enums generated by the script (which is tailored for v201 schemas) contain conversions that were fixed manually


ocpp/v201/call.py
```diff
@@ -0,0 +1,429 @@
+from typing import Dict, List
+from dataclasses import dataclass
+
+
+@dataclass
+class AuthorizeRequestPayload:
+    id_token: Dict
+    certificate: str = None
+    iso15118_certificate_hash_data: List = None
```
OrangeTux, line 9: Two comments that are meant for all classes in this file:

* Can you remove the term  'Request' from the class name.
* Can you fix the type hints for all attributes that could be `None` and use `Optional`? So `certificate: Optional[str] = None`.

ocpp/v201/enums.py
```diff
@@ -0,0 +1,1193 @@
...
+    AttributeEnumType is used by: Common:VariableAttributeType ,
+    getVariables:GetVariablesRequest.GetVariableDataType ,
+    getVariables:GetVariablesResponse.GetVariableResultType ,
+    setVariables:SetVariablesRequest.SetVariableDataType ,
+    setVariables:SetVariablesResponse.SetVariableResultType
+    """
+    actual = "Actual"
+    target = "Target"
+    minSet = "MinSet"
+    maxSet = "MaxSet"
```
OrangeTux, line 93: Can you use `snake_case` for attributes?

ocpp/v201/enums.py
```diff
@@ -0,0 +1,1193 @@
+class Action:
```
OrangeTux, line 1: Can you subclass from `str` and `Enum`? [As is done for the OCPP 1.6 enums?](https://github.com/mobilityhouse/ocpp/blob/master/ocpp/v16/enums.py)

OrangeTux (changes requested):

ocpp/v201/enums.py
tropxy: In snake_case, what would be correct in this situation:
```python
re.sub(r'(?<!^)(?=[A-Z])', '_', "ALBama").lower()
> 'a_l_bama'

or 

re.sub('([a-z0-9])([A-Z])', r'\1_\2', "ALABama").lower()
> 'alabama'
```

OrangeTux: `snake_case` separates different words or terms with `_`. The problem with `ALABama` is that it is a single word with weird capitalization. The correct outcome is `alabama`.

OrangeTux (approved):

PR #96, opened 2020-06-25
Create decorator

tmh-azinhal:


ocpp/charge_point.py
```diff
@@ -74,13 +74,39 @@ def remove_nones(dict_to_scan):
...
+    latency or not.
+
+    delay (int): average (?) delay to add to call results MILISECONDS
+    frequency (float): percentage of times the delay will be added [0.0, 1.0]
+    """
+    async def decorator(function):
+        async def wrapper(*args, **kwargs):
+            result = await function(*args, **kwargs)
+            """
+            import numpy as np
```
Roger, line 91: I think that making numpy as a dependency of this is way to heavy

ocpp/charge_point.py
```diff
@@ -220,6 +250,7 @@ async def _handle_call(self, msg):
             # when no '_on_after' hook is installed.
             pass
 
+    @add_latency #pass argument here to add the required latency, somehow
```
Roger, line 253: Since we depends on some inputs, I don't think that a decorator is the best approach, because is not explicit that you depend on the function to be a bounded method and that the class should contain this attributes that could colide, I think is better to create an async function and use it where we want to introduce the delay.

ocpp/charge_point.py
tmh-azinhal: Yes, I am aware. In no way was I thinking of adding numpy here. And I'm still thinking if I can find another way to not use it at all.

ocpp/charge_point.py
laysauchoa: why not sleep with async?

tmh-azinhal:
I discussed this with @mgonzalezperna and @tropxy and finally when I asked the correct question I figured out how to best approach this.

The problem was that, although `@decorator` and `fn = decorator(func)` are the same. When decorating a class' method with `@decorator` this will be done when the class is created. Thus, will not allow me to pass arguments to the decorator when I create an instance of the class - which is what I wanted.

I decided then to decorate the target method on the `init` of the class but this will hide the undecorated method from anyone using the class. Which is also not a behaviour that I want.

If possible I want to be able to decorate anything I want, and not specifically the `ChargePoint.call` method.

Since I am using VirtualChargePoint, which is a subclass of the ocpp ChargePoint, I can decorate the method call in there. And taking it a step further, I can decorate each call method from each *instance* of the VirtualChargePoint in the site simulator.

So that is the approach I will take. Not impacting the ocpp repo or the virtual-charge-point repo.

I hope that makes sense, @Roger and @laysauchoa.

PR #106, opened 2020-08-26
Allow __str__ representation to print exceptions

### LINKED ISSUE #105: Implement __str__ for all OCPP exceptions
Currently errors in `ocpp.exceptions` lack a `__str__()` implementation. Therefore the following code doesn't print the error:

``` python
try:
   // Something that raises an exception.
except ocpp.exceptions.OCPPError as e:
  logger.warning("Failed to do something: %s", e)
```

laysauchoa:
Issue description:

Currently, errors in ocpp.exceptions lack a __str__() implementation.
Therefore, the following code doesn't print the error:

```
try:
   // Something that raises an exception.
except ocpp.exceptions.OCPPError as e:
  logger.warning("Failed to do something: %s", e)
```

Closes #105

### DIFF
--- a/ocpp/exceptions.py
+++ b/ocpp/exceptions.py
@@ -25,8 +25,8 @@ class OCPPError(Exception):
                f" details={self.details}>"
 
     def __str__(self):
-        return f"<{self.__class__.__name__} - description={self.description},"\
-               f" details={self.details}>"
+        return f"{self.__class__.__name__}: {self.description},"\
+               f" {self.details}"
 
 
 class NotImplementedError(OCPPError):
--- a/ocpp/exceptions.py
+++ b/ocpp/exceptions.py
@@ -24,6 +24,10 @@ class OCPPError(Exception):
         return f"<{self.__class__.__name__} - description={self.description},"\
                f" details={self.details}>"
 
+    def __str__(self):
+        return f"<{self.__class__.__name__} - description={self.description},"\
+               f" details={self.details}>"
+
 
 class NotImplementedError(OCPPError):
     code = "NotImplemented"


ocpp/exceptions.py
```diff
@@ -24,6 +24,10 @@ def __repr__(self):
         return f"<{self.__class__.__name__} - description={self.description},"\
                f" details={self.details}>"
 
+    def __str__(self):
+        return f"<{self.__class__.__name__} - description={self.description},"\
+               f" details={self.details}>"
```
DColadas, line 29:
```suggestion
        return f"{self.__class__.__name__}: {self.description},"\
               f" {self.details}"
```

In my opinion, because this method would most likely be used to show information to the client (not for debugging purposes, as `__repr__` should be used for that), `__str__` shouldn't contain the `<` and `>` beginning and ending. Also, I'd not write `description=` and `details=` but directly print them.

Although nitpicky, I'd also say using `:` instead of `-` keeps the format more similar to how exceptions are printed by default. That's debatable, though.

DColadas (changes requested):
In my opinion, unlike `__repr__`, `__str__` shouldn't contain the `<` and `>` beginning and ending.

Also, I'd not write `description=` and `details=`. This method would most likely be used to show information to the client, not for debugging purposes (`__repr__` should be used for that).

ocpp/exceptions.py
laysauchoa: yeah, looks better actually. I was just not sure if the exceptions would include in the future details and descriptions.

OrangeTux (approved):

Issue #110, opened 2020-09-10
Subclass enums from both str and enum.Enum


OrangeTux:
Subclassing the enums from both `str` and `enum.Enum` gives the flexibility of `enum.Enum` in a backwards compatible way.

This ticket has been created as result from the discussion in #61

PR #112, opened 2020-09-11
Subclass OCPP 1.6 enums from str and enum.Enum

OrangeTux:
Subclassing from `enum.Enum` has several benefits:

* type hinting of handlers, Calls and CallResults can be improved
* improves integration with frame works like Django
* enums can be enumerated more easily

Subclassing from `str` gives backwards compatibility. Without it
all calls like `AuthorizationStatus.accepted` had to be replaced with
`AuthorizationStatus.accepted.value()`.

Fixes: #110


tropxy (dismissed):
👍

tropxy (approved):
👍 👍

Issue #104, opened 2020-08-23
Implement way to allow users to handle CallErrors


OrangeTux:
Currently `ChargePoint.call()` returns `None` when the remote end returns a `CallError`. That is because [`ChargePoint.call()`](https://github.com/mobilityhouse/ocpp/blob/master/ocpp/charge_point.py#L258-L260) silences the `CallError`.

This is problematic when one wants to act upon receiving a `CallError`. 

I'm not sure yet how and where to change the code so one can handle `CallError`s. The easiest way is to change the body of `ChargePoint.call()` to this:

``` python
if response.message_type_id == MessageType.CallError:
    raise response.to_exception()  
```

But this change is not backwards compatible. In existing code bases one users of this library might need to wrap all there calls to `ChargePoint.call()` inside try/except to avoid their code from breaking, like:

``` python
try:
   self.call(call.HeartbeatPayload())
except ocpp.exceptions.OCPPError as e:
   pass
```

OrangeTux:
In order to main backwards compatibility I suggest to add a keyword argument `suppress` to `ChargePoint.call()` that defaults to `True.

tropxy:
Fixed in Raise exception if response from call() is CallError #124

PR #124, opened 2020-10-07
Raise exception if response from call() is CallError

tmh-azinhal:
This PR references issue #104 .

If the response is a CallError, an exception will be raised instead of being silenced.
To maintain backwards compatibility a `suppress=True` keyword argument was added in `Chargepoint.call`.


ocpp/charge_point.py
```diff
@@ -224,7 +224,7 @@ async def _handle_call(self, msg):
             # when no '_on_after' hook is installed.
             pass
 
-    async def call(self, payload):
+    async def call(self, payload, surpress=True):
```
OrangeTux, line 227: s/surpress/suppress

OrangeTux: Can you also update the docstring and mention this new argument?

OrangeTux (changes requested):
You can add a test? I'll come back later to

ocpp/charge_point.py
tropxy: I would change the name to "suppress_ocpp_exception" or "suppress_exception". 

Or later on in the code I dont remember what the suppress is suppressing

tropxy (changes requested):

ocpp/charge_point.py
tmh-azinhal: I would say that keeping the name simple would be preferable. I hope the docstring explains what the argument's purpose is.

tropxy: ok, fair enough

OrangeTux:
I promised some advice when it comes to writing tests for the piece of code you introduced.

In [tests/v16/test_v16_charge_point.py] you can add the tests. For the test you can reuse the `base_central_system` as is done in other tests in that file.

This test require a call, that is the request, and a call error. The latter is the response. Because of limitations of testing you first need prepare the response.  Create an instance of `ocpp.messages.CallError`, turn it into json and pass it to `base_central_system.route_message()`. [`route_message()`](https://github.com/mobilityhouse/ocpp/blob/master/ocpp/charge_point.py#L128) will put the response on [an internal response queue](https://github.com/mobilityhouse/ocpp/blob/master/ocpp/charge_point.py#L148).

Now you can create the request. You need to instantiate a payload from `call.v16.call` and 'send' that to the other end of the connection using `base_central_system.call()`. [`call()`](https://github.com/mobilityhouse/ocpp/blob/master/ocpp/charge_point.py#L227) will [use the response queue](https://github.com/mobilityhouse/ocpp/blob/master/ocpp/charge_point.py#L259-L260) to look for response to its request.

Note that it's important that the request and response have [the same unique id](https://github.com/mobilityhouse/ocpp/blob/master/ocpp/charge_point.py#L290-L291). Otherwise a response can't be matched to a request.

I hope this is enough information. If not, let me know.

OrangeTux (approved):

OrangeTux (changes requested):
The lint job on CircleCI compains. Can you make the job happy?

OrangeTux (approved):
Thanks!

tropxy (approved):

PR #126, opened 2020-10-11
[WIP] Handle call error #104

tropxy:
This integrates the possibility to handle CallError messages without breaking backwards compatibility.

The implementor can decide on what todo with the Error message. 

## Central System code

```python
...

class ChargePoint(cp):
    @on(Action.BootNotification)
    def on_boot_notitication(self, charge_point_vendor: str, charge_point_model: str, **kwargs):
        raise InternalError("Autobots Rollout Failed")
```

## Charge Point Code
```python
...

class ChargePoint(cp):
    async def send_boot_notification(self):
        request = call.BootNotificationPayload(
            charge_point_model="Optimus",
            charge_point_vendor="The Mobility House"
        )

        response = await self.call(request)
            print(response)
            print("Error Code: ", response.error_code)
            print("Exception: ", response.to_exception())
            return
        if response.status == RegistrationStatus.accepted:
            print("Connected to central system.")
        else:
            print(response)
```

### DIFF
--- a/ocpp/charge_point.py
+++ b/ocpp/charge_point.py
@@ -261,7 +261,7 @@ class ChargePoint:
 
         if response.message_type_id == MessageType.CallError:
             LOGGER.warning("Received a CALLError: %s'", response)
-            return
+            return response
         else:
             response.action = call.action
             validate_payload(response, self._ocpp_version)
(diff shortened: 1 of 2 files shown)


tropxy:
After an internal discussion and given there is already a good approach to make this backwards comp, I decided to close this.

An issue soon will be opened with backwards compatible changes that we will want to undo before releasing version 1.0.0

PR #132, opened 2020-10-14
Fix 'Optional' type hints

### LINKED ISSUE #127: Fix type hints for attributes that are optional
Currently the classes as defined in `ocpp.*.call` and `ocpp.*.call_result` have optional attributes. But the type hints don't always reflect that.

For example:

``` python
@dataclass
class GetConfigurationPayload:
    key: List = None
```

This should become:

``` python
@dataclass
class GetConfigurationPayload:
    key: Optional[List] = None
```

OrangeTux:
Fixes #127

### DIFF
--- a/ocpp/v16/call.py
+++ b/ocpp/v16/call.py
@@ -57,10 +57,10 @@ class ClearCachePayload:
 
 @dataclass
 class ClearChargingProfilePayload:
-    id: int = None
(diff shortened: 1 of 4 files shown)


tropxy (approved):

laysauchoa (approved):
:tada:

PR #134, opened 2020-10-15
Auto generate enums for v201

### LINKED ISSUE #40: Auto Generate Enums Class based on json schemas
The idea is to follow the already implemented generator for the call and call result payloads and generate classes for all the OCPP spec enums. After a quick research, I found that there are schemas like the Firmware Status Notification: [https://github.com/mobilityhouse/ocpp/blob/e30a0db5a8afaad24ff9612bad0208381a3af2de/ocpp/v20/schemas/FirmwareStatusNotificationRequest_v1p0.json#L7](https://github.com/mobilityhouse/ocpp/blob/e30a0db5a8afaad24ff9612bad0208381a3af2de/ocpp/v20/schemas/FirmwareStatusNotificationRequest_v1p0.json#L7)

that does not include a definition of the Enumeration and just adds the enums to the property itself which is not the way used in other schemas like in GetVariablesRequest [https://github.com/mobilityhouse/ocpp/blob/e30a0db5a8afaad24ff9612bad0208381a3af2de/ocpp/v20/schemas/GetVariablesRequest_v1p0.json#L4](https://github.com/mobilityhouse/ocpp/blob/e30a0db5a8afaad24ff9612bad0208381a3af2de/ocpp/v20/schemas/GetVariablesRequest_v1p0.json#L4
)

First, we can build a script to generate Enums based on the schemas like the GetVariablesRequest one and see which ones were generated and if there are Enums missing

tropxy:
fixes #40

### DIFF
--- a/ocpp/charge_point.py
+++ b/ocpp/charge_point.py
@@ -261,7 +261,7 @@ class ChargePoint:
 
         if response.message_type_id == MessageType.CallError:
             LOGGER.warning("Received a CALLError: %s'", response)
-            return response
+            return
         else:
             response.action = call.action
             validate_payload(response, self._ocpp_version)
--- a/ocpp/charge_point.py
+++ b/ocpp/charge_point.py
@@ -261,7 +261,7 @@ class ChargePoint:
 
         if response.message_type_id == MessageType.CallError:
             LOGGER.warning("Received a CALLError: %s'", response)
-            return
+            return response
         else:
             response.action = call.action
             validate_payload(response, self._ocpp_version)
--- a/scripts/schema_to_enums_v201.py
+++ b/scripts/schema_to_enums_v201.py
@@ -99,8 +99,8 @@ if __name__ == '__main__':
     # The former contains the enum objects that are later parsed to the
     # enum.py, using the str representation. And the latter, is used to keep
     # track of the already processed Enums, avoiding data duplication
-    for schema_ in schemas:
-        parse_schema(schema_)
+    for schema in schemas:
+        parse_schema(schema)
 
     with open('enums.py', 'wb+') as f:
         f.write(b"from enum import Enum\n")
--- a/scripts/schema_to_enums_v201.py
+++ b/scripts/schema_to_enums_v201.py
@@ -1,0 +1,106 @@
+#!/usr/bin/env python
+import sys
+import re
+import json
+from pathlib import Path
+
+
+def create_class(name):
+    return normalclass(name)
+
+
+def create_attribute(name, casing_style="snake_case"):
+    # Assumes the variable name is a concatenation of capitalized therms, e.g.,
+    # "CamelCase" or "SuperHyperMega"
+    # Convert to CamelCase
+    name_normalized = re.sub('[-.]', '_', name)
+    if casing_style == "CamelCase":
+        name_converted = name_normalized[0].lower() + name_normalized[1:]
+    else:
+        name_converted = re.sub('([a-z0-9])([A-Z])', r'\1_\2',
+                                name_normalized).lower()
+
+    return attribute(name, name_converted)
+
+
+class normalclass:
+    def __init__(self, name):
+        self.name = name
+        self.attrs = []
+
+    def add_attr(self, attr):
+        self.attrs.append(attr)
+
+    def __str__(self):
+        output = f"class {self.name}(str, Enum):\n"
+
+        if len(self.attrs) == 0:
+            return output + "    pass\n"
+
+        for attr in self.attrs:
+            output += str(attr)
+
+        return output
+
+
+class attribute:
+    def __init__(self, name, name_converted):
+        self.name = name
+        self.name_converted = name_converted
+
+    def __str__(self):
+        name = self.name
+        if not re.match("^[a-zA-Z_]", self.name):
+            name = "_" + self.name
+
+        definition = f'    {self.name_converted} = "{name}"\n'
+
+        return definition
+
+    def __repr__(self):
+        return f"<{self.name}, {self.name_converted}> "
+
+
+
+enum_types = []
+enum_types_names = []
+
+
+def parse_schema(schema):
+    with open(schema, "r") as f:
+        schema = json.loads(f.read())
+
+    try:
+        definitions = schema['definitions']
+    except KeyError:
+       pass
+
+    for enum_type, value in definitions.items():
+        if "Enum" in enum_type:
+            type_name = enum_type.replace("Enum", '')
+            if type_name not in enum_types_names:
+                nc = create_class(type_name)
+                for enum_attr in value['enum']:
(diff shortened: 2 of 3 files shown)


scripts/schema_to_enums_v201.py
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env python
+import sys
+import re
+import json
+from pathlib import Path
+
+
+def create_class(name):
+    return normalclass(name)
```
OrangeTux, line 9: I'm not sure why this function exists. Why not call the constructor of `normalclass` directly?

scripts/schema_to_enums_v201.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+def create_attribute(name, casing_style="snake_case"):
+    # Assumes the variable name is a concatenation of capitalized therms, e.g.,
```
OrangeTux, line 13: s/therms/terms

scripts/schema_to_enums_v201.py
```diff
@@ -0,0 +1,106 @@
...
+    if casing_style == "CamelCase":
+        name_converted = name_normalized[0].lower() + name_normalized[1:]
+    else:
+        name_converted = re.sub('([a-z0-9])([A-Z])', r'\1_\2',
+                                name_normalized).lower()
+
+    return attribute(name, name_converted)
+
+
+class normalclass:
```
OrangeTux, line 26: s/normalclass/NormalClass

scripts/schema_to_enums_v201.py
```diff
@@ -0,0 +1,106 @@
...
+        if len(self.attrs) == 0:
+            return output + "    pass\n"
+
+        for attr in self.attrs:
+            output += str(attr)
+
+        return output
+
+
+class attribute:
```
OrangeTux, line 46: s/attribute/Attribute

scripts/schema_to_enums_v201.py
```diff
@@ -0,0 +1,106 @@
...
+        self.name_converted = name_converted
+
+    def __str__(self):
+        name = self.name
+        if not re.match("^[a-zA-Z_]", self.name):
+            name = "_" + self.name
+
+        definition = f'    {self.name_converted} = "{name}"\n'
+
+        return definition
```
OrangeTux, line 58: This can shortened to a single line.

scripts/schema_to_enums_v201.py
```diff
@@ -0,0 +1,106 @@
...
+                enum_types_names.append(nc.name)
+
+
+if __name__ == '__main__':
+    if len(sys.argv) != 2:
+        print("Pass path to folder with schemas")
+        sys.exit(-1)
+
+    p = Path(sys.argv[1])
+    schemas = list(p.glob("*.json"))
```
OrangeTux, line 96: Can you document the arguments?

scripts/schema_to_enums_v201.py
```diff
@@ -0,0 +1,106 @@
...
+
+    for schema in schemas:
+        parse_schema(schema)
+
+    with open('enums.py', 'wb+') as f:
```
OrangeTux, line 101: IIRC `+` will open the file in a mode where `write()` appends to the file instead of overwriting the original source. So if you run this script twice the output has all enum definitions twice. I'm not sure if you want that.

OrangeTux (changes requested):
Can you add a bit more documentation? A script like won't be used often. So if we've to use or modify it in the future it's important that documentation explains how the code works.

scripts/schema_to_enums_v201.py
tropxy: Legacy code from the other script

scripts/schema_to_enums_v201.py
tropxy: No, that would be the case if flag 'a' would be used... 'wb+' writes as binary and truncate the file

ocpp/charge_point.py
```diff
@@ -261,7 +261,7 @@ async def call(self, payload):
 
         if response.message_type_id == MessageType.CallError:
             LOGGER.warning("Received a CALLError: %s'", response)
-            return
+            return response
```
OrangeTux: This shouldn't be part of this PR. It's already worked on in a different PR>

scripts/schema_to_enums_v201.py
```diff
@@ -92,15 +91,21 @@ def parse_schema(schema):
...
     schemas = list(p.glob("*.json"))
 
-    for schema in schemas:
-        parse_schema(schema)
+    # This loop will update the lists enum_types and enum_types_names:
+    # The former contains the enum objects that are later parsed to the
+    # enum.py, using the str representation. And the latter, is used to keep
+    # track of the already processed Enums, avoiding data duplication
+    for schema_ in schemas:
+        parse_schema(schema_)
```
OrangeTux, line 103: Can't you use `schema` over `schema_`?

scripts/schema_to_enums_v201.py
```diff
@@ -4,26 +4,29 @@
...
 from pathlib import Path
 
-
-def create_class(name):
-    return normalclass(name)
+enum_types = []
+enum_types_names = []
 
 
 def create_attribute(name, casing_style="snake_case"):
```
OrangeTux, line 11: The value of `snake_case` is never used. Wouldn't a boolean more appropriate?

OrangeTux (changes requested):

scripts/schema_to_enums_v201.py
tropxy: I was layering ground for more possible case styles. You prefer me to do that change and make it bool? or include it in a if/elif  and even raise an error in case no answer is provided?

scripts/schema_to_enums_v201.py
tropxy: My linter complains because schema shadows the use of schema in parse_schema

OrangeTux: Isn't that a problem with your linter? The scopes of both variables are totally different. `schema_` lives inside the main, while `schema` lives inside `parse_schema()`

tropxy: You are right

scripts/schema_to_enums_v201.py
tropxy: Let me know asap

OrangeTux: > I was layering ground for more possible case styles. 

When do you expect changes to output casing? The casing of Python code in our library is static, either snake_case or CamelCase. I don't see that changing soon.

tropxy: Ok, agreed.

OrangeTux (approved):

PR #135, opened 2020-10-15
Fix deadlock in `@after` handler

### LINKED ISSUE #130: `@after()` handler is causing a deadlock
Hello,

I'm using this project to implement a Central System to test a (proprietary) Charge Point implementation.
In order to simulate several behaviors, I want to trigger CS initiated requests, but I can't get it to work.

For example, let's imagine you want to trigger a ChangeConfiguration request (from the CS to the CP) after a Heartbeat request (from the CP to the CS).

My naive attempt is to include the following code in the CS sample class (OCPP 1.6). But the execution hangs after the call().
The PDU is sent to the CP, but no further data is received.

```
@after(Action.Heartbeat)
async def update_heartbeat_interval(self):
	request = call.ChangeConfigurationPayload(
		key="HeartbeatInterval",
		value="123"
	)
	response = await self.call(request)
	#do something with the response...(code never reached)
```

I'm new to Python, and I don't know if I understand the project architecture or coroutines correctly.
Thanks to all contributors for this project.

OrangeTux:
0.7.1 reintroduced a deadlock bug that was already fixed in 0.4.2.

Sending a call in an `@after` handler would cause a deadlock.

Fixes #130

### DIFF
--- a/ocpp/charge_point.py
+++ b/ocpp/charge_point.py
@@ -218,7 +218,7 @@ class ChargePoint:
             # after handler
             response = handler(**snake_case_payload)
             if inspect.isawaitable(response):
-                await response
+                asyncio.ensure_future(response)
         except KeyError:
             # '_on_after' hooks are not required. Therefore ignore exception
             # when no '_on_after' hook is installed.


laysauchoa (approved):
sgtm!

ocpp/charge_point.py
```diff
@@ -218,7 +218,7 @@ async def _handle_call(self, msg):
             # after handler
             response = handler(**snake_case_payload)
             if inspect.isawaitable(response):
-                await response
+                asyncio.ensure_future(response)
```
tropxy, line 221: cant we use the higher level API (asyncio.create_task)?

tropxy (changes requested):

ocpp/charge_point.py
OrangeTux: Good suggestion. Unfortunately  `asyncio.create_task()` was added in Python 3.7, while we also support Python 3.6.

tropxy: True. Then this is resolved

tropxy (approved):

PR #145, opened 2020-10-22
Added Example to OCPP 2.0.1

### LINKED ISSUE #142: Create examples for OCPP 2.0.1

tropxy:
Also fixed subprotocol handling in all examples

fixes #142

### DIFF
--- a/examples/v16/central_system.py
+++ b/examples/v16/central_system.py
@@ -34,8 +34,13 @@ async def on_connect(websocket, path):
     """ For every new charge point that connects, create a ChargePoint
     instance and start listening for messages.
     """
-    requested_protocols = websocket.request_headers['Sec-WebSocket-Protocol']
-    if bool(websocket.subprotocol):
+    try:
+        requested_protocols = websocket.request_headers[
+            'Sec-WebSocket-Protocol']
+    except KeyError:
+        logging.info("Client hasn't requested any Subprotocol. "
+                     "Closing Connection")
+    if websocket.subprotocol:
         logging.info("Protocols Matched: %s", websocket.subprotocol)
     else:
         # In the websockets lib if no subprotocols are supported by the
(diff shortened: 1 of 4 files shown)


examples/v16/central_system.py
```diff
@@ -31,10 +31,22 @@ def on_boot_notitication(self, charge_point_vendor: str, charge_point_model: str
 
 
 async def on_connect(websocket, path):
-    """ For every new charge point that connects, create a ChargePoint instance
-    and start listening for messages.
-
+    """ For every new charge point that connects, create a ChargePoint
+    instance and start listening for messages.
     """
+    requested_protocols = websocket.request_headers['Sec-WebSocket-Protocol']
```
OrangeTux, line 37: The bounds of `websocket.request_headers` aren't checked here. A client that doesn't send the header 'Sec-WebSocket-Protocol' will cause a `KeyError` here.

examples/v16/central_system.py
```diff
@@ -31,10 +31,22 @@ def on_boot_notitication(self, charge_point_vendor: str, charge_point_model: str
...
+    if bool(websocket.subprotocol):
+        logging.info("Protocols Matched: %s", websocket.subprotocol)
```
OrangeTux, line 39: Wouldn't `if websocket.subprotocol == '2.0.1` be more explicit?

OrangeTux (changes requested):

examples/v16/central_system.py
tropxy: For this example in specific, where I had used just one supported lib, yes. But in general, you can support several, so if the result of the subprotocol is positive, then you can continue with the communication, regardless of which subprotocol of the list you have chosen

examples/v16/central_system.py
tropxy: Agreed 👍

examples/v16/central_system.py
OrangeTux: Ok. Fair point.

PR #149, opened 2020-10-28
Converted all camelCase and PascalCase enums to snake_case, except Action enumeration

### LINKED ISSUE #114: Make casing of enum members consistent
The casing for members of enums in `ocpp.v16.enums` is inconsistent. 

It contains PascalCase...

``` python
class Action:
    Authorize = "Authorize"
```

...camelCase...

``` python
class ChargePointErrorCode:
    connectorLockFailure = "ConnectorLockFailure"
```

... and lowercase

``` python
class ChargingProfilePurposeType(str, Enum):
    chargepointmaxprofile = "ChargePointMaxProfile"
```

I suggest to use snake_case for all members. To maintain backwards compatibility I suggest to mark all non_snake_case members with a deprecation warning.

tropxy:
The conversion in the ocpp1.6 includes also the old enums to ensure backwards compatibility 

fixes #114

### DIFF
--- a/ocpp/v16/enums.py
+++ b/ocpp/v16/enums.py
@@ -42,7 +42,7 @@ class AuthorizationStatus(str, Enum):
     blocked = "Blocked"
     expired = "Expired"
     invalid = "Invalid"
-    concurrenttx = "ConcurrentTx"
+    concurrent_tx = "ConcurrentTx"
 
 
 class AvailabilityStatus(str, Enum):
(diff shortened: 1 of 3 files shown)


laysauchoa (commented):
looks good! but should this be done before these changes be merged? _to maintain backwards compatibility I suggest to mark all non_snake_case members with a deprecation warning._

OrangeTux:
I think we should maintain backward compatibility, at least for a while. Although I initially wrote mark deprecated attributes as such I'm unsure how to do that. Class attributes can't be decorated with `@deprecation`.

Also you forgot to update the OCPP 2.0.1 enums.

tropxy:
> I think we should maintain backward compatibility, at least for a while. Although I initially wrote mark deprecated attributes as such I'm unsure how to do that. Class attributes can't be decorated with `@deprecation`.
> 
> Also you forgot to update the OCPP 2.0.1 enums.

We have agreed a few days ago to include non compatible changes until 1.0.0, right? that is why I didnt cover that issue.
If we want to include backwards compatibility, then, we can simply include both ways of calling the enums.

> Also you forgot to update the OCPP 2.0.1 enums.
When 2.0.1 PR was created, was already with snake_case, so in this case there is no backward compatibility option

OrangeTux:
> We have agreed a few days ago to include non compatible changes until 1.0.0, right?

My understanding of the conversation is a bit different. It's true that SemVer allows to do breaking changes before the 1.0.0 release. So we could introduce breaking changes. But, still I think we need to careful with introducing breaking changes. It can break the applications of everyone who's relying on this packag. Even our own app would break. Therefore is suggest also keep the old attributes.

> When 2.0.1 PR was created, was already with snake_case, so in this case there is no backward compatibility option

Than I found a couple of bugs there:

https://github.com/mobilityhouse/ocpp/blob/master/ocpp/v201/enums.py#L106
https://github.com/mobilityhouse/ocpp/blob/master/ocpp/v201/enums.py#L323
https://github.com/mobilityhouse/ocpp/blob/master/ocpp/v201/enums.py#L442
https://github.com/mobilityhouse/ocpp/blob/master/ocpp/v201/enums.py#L550
https://github.com/mobilityhouse/ocpp/blob/master/ocpp/v201/enums.py#L653
https://github.com/mobilityhouse/ocpp/blob/master/ocpp/v201/enums.py#L664
https://github.com/mobilityhouse/ocpp/blob/master/ocpp/v201/enums.py#L1185

tropxy:
> > We have agreed a few days ago to include non compatible changes until 1.0.0, right?
> 
> My understanding of the conversation is a bit different. It's true that SemVer allows to do breaking changes before the 1.0.0 release. So we _could_ introduce breaking changes. But, still I think we need to careful with introducing breaking changes. It can break the applications of everyone who's relying on this packag. Even our own app would break. Therefore is suggest also keep the old attributes.
> 
> > When 2.0.1 PR was created, was already with snake_case, so in this case there is no backward compatibility option
> 
> Than I found a couple of bugs there:
> 
> https://github.com/mobilityhouse/ocpp/blob/master/ocpp/v201/enums.py#L106
> https://github.com/mobilityhouse/ocpp/blob/master/ocpp/v201/enums.py#L323
> https://github.com/mobilityhouse/ocpp/blob/master/ocpp/v201/enums.py#L442
> https://github.com/mobilityhouse/ocpp/blob/master/ocpp/v201/enums.py#L550
> https://github.com/mobilityhouse/ocpp/blob/master/ocpp/v201/enums.py#L653
> https://github.com/mobilityhouse/ocpp/blob/master/ocpp/v201/enums.py#L664
> https://github.com/mobilityhouse/ocpp/blob/master/ocpp/v201/enums.py#L1185

Fair enough, I will address your points

OrangeTux (approved):
:+1:

Issue #154, opened 2020-11-13
Cache `Draft4Validator` instead of dict version of JSON schema.


OrangeTux:
I've done some micro benchmarks of OCPP and found that a few hot spots that are easy to optimize.

Currently OCPP caches the JSON schema as a Python dict. That prevents  reading to disk and parsing the schema if a schema is used more than one time. But still every time a schema is used this dict has to be turned into a [`Draft4Validator`](https://github.com/Julian/jsonschema/blob/main/jsonschema/validators.py#L462) instance

Caching the `Draft4Validator` instance reduces the time to handle a `Call` significantly.

I've used the following script to benchmark. 
It requires [`pytest-benchmark`](https://pypi.org/project/pytest-benchmark/). This package doesn't support async functions natively, but @mbello provided some code to fix that in [this issue](https://github.com/ionelmc/pytest-benchmark/issues/66).

The script can be executed using `

``` bash
$ pytest tests/test_benchmark.py
```

``` python
# tests/benchmark.py
import asyncio
import threading
import pytest
from datetime import datetime

from ocpp.routing import on
from ocpp.v16 import ChargePoint as cp, call_result
from ocpp.v16.enums import Action, RegistrationStatus
from ocpp.messages import Call


@pytest.fixture
def boot_notification_call():
    return Call(
        unique_id="1",
        action=Action.BootNotification,
        payload={
            'chargePointVendor': 'Alfen BV',
            'chargePointModel': 'ICU Eve Mini',
            'firmwareVersion': "#1:3.4.0-2990#N:217H;1.0-223",
        }).to_json()


@pytest.fixture
def meter_values_call():
    return Call(
        unique_id="1",
        action=Action.MeterValues,
        payload={
            'transactionId': 6998264,
            'connectorId': 1,
            'meterValue': [{
                'timestamp': '2020-07-09T12:52:40Z',
                'sampledValue': [{
                    'location': 'Outlet',
                    'measurand': 'Energy.Active.Import.Register',
                    'unit': 'Wh',
                    'value': '419780.000'
                }, {
                    'context': 'Sample.Periodic',
                    'location': 'Outlet',
                    'measurand': 'Energy.Active.Import.Register',
                    'phase': 'L1',
                    'unit': 'Wh',
                    'value': '242730.000'
                }, {
                    'context': 'Sample.Periodic',
                    'location': 'Outlet',
                    'measurand': 'Energy.Active.Import.Register',
                    'phase': 'L2',
[...]

PR #155, opened 2020-11-13
Optimize validation by caching `Draft4Validator`s

OrangeTux:
Creating a `Draft4Validator` instance from a dict representation of a
JSON schema is pretty expensive. Therefore caching the instances makes
sense.

I've done some micro benchmarks. Below are the results:

before
Screenshot_2020-11-13_17-13-16

As expected benchmarks `test_xxxx_with_validation` show an significant performance improvement. 
`test_meter_values_with_validation` is 1.88 times faster.
`test_boot_notifications_with_validation` is 2.60  times faster.

Fixes: #154

### DIFF
--- a/ocpp/messages.py
+++ b/ocpp/messages.py
@@ -170,12 +170,13 @@ def get_validator(
     if ocpp_version == "2.0":
         schema_name += '_v1p0'
 
+    cache_key = schema_name + ocpp_version
+    if cache_key in _schemas:
+        return _schemas[cache_key]
+
     dir,  _ = os.path.split(os.path.realpath(__file__))
     relative_path = f'{schemas_dir}/schemas/{schema_name}.json'
     path = os.path.join(dir, relative_path)
-
-    if relative_path in _validators:
-        return _validators[relative_path]
 
     # The JSON schemas for OCPP 2.0 start with a byte order mark (BOM)
     # character. If no encoding is given, reading the schema would fail with:
@@ -184,9 +185,9 @@ def get_validator(
     with open(path, 'r', encoding='utf-8-sig') as f:
         data = f.read()
         validator = Draft4Validator(json.loads(data, parse_float=parse_float))
-        _validators[relative_path] = validator
+        _validators[cache_key] = validator
 
-    return _validators[relative_path]
+    return _validators[cache_key]
 
 
 def validate_payload(message, ocpp_version):
--- a/ocpp/messages.py
+++ b/ocpp/messages.py
@@ -174,7 +174,7 @@ def get_validator(
     relative_path = f'{schemas_dir}/schemas/{schema_name}.json'
     path = os.path.join(dir, relative_path)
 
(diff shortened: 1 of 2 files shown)


ocpp/messages.py
```diff
@@ -101,6 +104,10 @@ def get_schema(message_type_id, action, ocpp_version, parse_float=float):
     is used to parse floats. It must be a callable taking 1 argument. By
     default it is `float()`, but certain schema's require `decimal.Decimal()`.
     """
+    warnings.warn(
+        "Depricated as of 0.8.1. Please use `ocpp.messages.get_validator()`."
```
Roger, line 108: typo, it's "Deprecated"

Roger (commented):
Looks really good, just a small typo there

ocpp/messages.py
```diff
@@ -134,6 +141,55 @@ def get_schema(message_type_id, action, ocpp_version, parse_float=float):
...
+    if message_type_id == MessageType.CallResult:
+        schema_name += 'Response'
+    elif message_type_id == MessageType.Call:
+        if ocpp_version in ["2.0", "2.0.1"]:
+            schema_name += 'Request'
+
+    if ocpp_version == "2.0":
+        schema_name += '_v1p0'
+
+    cache_key = schema_name + ocpp_version
```
proelke, line 173: One small suggestion for readability/consistency would be to put an underscore or something to separate the name and version. Not a big deal either way though.

proelke (dismissed):
Made one small comment but otherwise looks good to me. Good work finding/testing this.

ocpp/messages.py
```diff
@@ -134,6 +141,55 @@ def get_schema(message_type_id, action, ocpp_version, parse_float=float):
...
+    if cache_key in _schemas:
+        return _schemas[cache_key]
```
tropxy, line 175: was it not supposed to be:
```python
if cache_key in _validators:
        return _validators[cache_key]
```

tropxy (changes requested):

ocpp/messages.py
OrangeTux: Thanks for the feedback. I changed it.

ocpp/messages.py
OrangeTux: You're right. Thanks for notificing.

proelke (approved):

ocpp/messages.py
tropxy: He has deprecated your comment @Roger :P

tropxy: @OrangeTux fix that tiny typo so that I can approve this peral

proelke: "peral"

tropxy: hahaha

tropxy: that was the typo, I was talking about ;)

tropxy (approved):

PR #159, opened 2020-11-30
Improve documentation of on()

adamchainz:
Cover that it can take an async function as well as a sync one, and use that in examples. Using async would be preferred as it allows `await`-ing DB I/O etc. whilst handling a call.

Also make the terms used a bit more consistent - “handler” throughout instead of “hook”, and recommend taking `**kwargs`` to ignore future arguments.


OrangeTux (approved):
Thanks!

Issue #167, opened 2021-01-18
OCPP 2.0.1 call payloads RequestStartTransactionPayload and RequestStopTransactionPayload are missing 


henbier:
Hi,

it seems like call.py does **NOT** have a **Request**StartTransactionPayload or a **Request**StopTransactionPayload, just a [StartTransactionPayload](https://github.com/mobilityhouse/ocpp/blob/master/ocpp/v201/call.py#L291) and a [StopTransactionPayload](https://github.com/mobilityhouse/ocpp/blob/master/ocpp/v201/call.py#L300). 
There **IS** a schema for [**Request**StartTransactionPayload](https://github.com/mobilityhouse/ocpp/blob/master/ocpp/v201/schemas/RequestStartTransactionRequest.json) and [**Request**StopTransactionPayload](https://github.com/mobilityhouse/ocpp/blob/master/ocpp/v201/schemas/RequestStopTransactionRequest.json) but **NOT** for StartTransactionPayload or StopTransactionPayload.
[call_result.py]() however does have a [**Request**StartTransactionPayload](https://github.com/mobilityhouse/ocpp/blob/master/ocpp/v201/call_result.py#L255) & [**Request**StopTransactionPayload](https://github.com/mobilityhouse/ocpp/blob/master/ocpp/v201/call_result.py#L262). 

I linked the exact lines in call.py and call_result.py along with the schemas I'm referring to.

_This leads to the following problem._ 
Using call.**Request**StartTransactionPayload(...) yields an _AttributeError_, because there is no function called **Request**StartTransactionRequest. (Code Example 1)
Using call.StartTransactionPayload(...) yields an _occp.exceotions.ValidationsError_, because there is no schema for StartTransactionRequest. (Code Example 2)

Code Example 1:
```
request = call.RequestStartTransactionPayload(
        request_start_id=some_id,
        id_token=some_token
)
-----------------------------------------------------
AttributeError: module 'ocpp.v201.call' has no attribute 'RequestStartTransactionPayload'
```

Code Example 2:
```
request = call.StartTransactionPayload(
        request_start_id=self.transaction_id,
        id_token=self.id_token
)
-----------------------------------------------------
ocpp.exceptions.ValidationError: Failed to load validation schema for action 'StartTransaction': [Errno 2] No such file or directory: 'C:\\...\\ocpp\\v201/schemas/StartTransactionRequest.json'
```

### **Same goes for RequestStopTransactionRequest.**
[...]

PR #169, opened 2021-01-19
Fix OCPP 2.0.1 call payloads RequestStartTransactionPayload and RequestStopTransactionPayload

OrangeTux:
The prefix `Request` was missing for the OCPP 2.0.1 call payloads
`RequestStartTransaction` and `RequestStopTransaction`.

Fix: #167

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change log
 
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


laysauchoa (approved):
nice! good that you found out :+1:

tropxy:
I know why this error appeared. The schemas_to_dataclass script outputs the class names with "RequestPayload" at the end. By the time I just Substitute the word Request by nothing, to remove it and doing so, as a side effect, also removed the Request prefix from those two classes :/

Issue #168, opened 2021-01-19
ocpp library as an event based message handler


villekr:
Enhancement issue based on original discussion topic: https://github.com/mobilityhouse/ocpp/discussions/165.

Main motivation for this is to:
- Allow ocpp library usage without tight decoupling with websocket
- Introduce ASGI compliant interface for http and websocket messages

villekr:
Current branch in PR https://github.com/mobilityhouse/ocpp/pull/171 supports now all three versions:
1. Current way of bundling both websocket connection and routes within one class, which derives from version specific ChargePoint classes (see examples/v16/central_system.py)
2. (New) stateful event based way where Central System holds Charging Station instances and based on id routes messages to correct instance.
3. (New) stateless event based way where messages are routed directly to correct action handler and it's up to action handler to figure out based on context-argument charging_station related information.

Both new ways implement ASGI interface and therefor must be run behind ASGI server (e.g. uvicorn).

villekr:
I've been lately too busy on other work so haven't been able to proceed any further with this. However, my plan is to next try to make a "ASGI first" version and then try to think how current ocpp implementation would fit to that model.

villekr:
Quite a much time has passed since I had time to proceed with this change. But now I have a next iteration draft ready. Let me first review what has been done so far:

- The first iteration was a prototype event based approach (to be able to run ocpp central system as serverless). Code is here: https://github.com/villekr/ocpp-serverless-example. Conclusion: serverless approach seems doable but code was just sort of hack.
- The second iteration was an attempt to modify ocpp library so that it would support both current ChargePoint based and event based stateless approaches. Code is here: https://github.com/mobilityhouse/ocpp/pull/171. Conclusion: the observation is that event based approach can't be implemented without braking compatibility with the ChargePoint based approach. Also the library interface and usage will get a bit confusing when there are two ways how library can be used.
[...]

PR #171, opened 2021-01-24
ocpp library as an event based message handler #168

villekr:
Work in progress related to issue #168.

This change adds a new Router base class that contains message routing functionality and modifies it to allow stateless and event based message handling. ChargePoint inherits this new base class and works unchanged providing a way to implement stateful bundled (i.e. both action handlers and websocket connection must be within ChargePoint class) message handler.

A new Router subclass in asgi_routing.py implements ASGI interface and enables decoupling action handlers from websocket connection.


ocpp/asgi_routing.py
```diff
@@ -0,0 +1,141 @@
...
+    async def _http_handler(self, scope, receive, send):
+        # ASGI http.request event type handling can be extended in subclasses
+        # as interface between "Service" and "ASGI server" depends about
+        # the used service (e.g. AWS).
+        pass
+
+    async def on_connect(self, context):
+        if not self._stateless:
+            # Actual implementation for stateful router missing
+            raise NotImplementedError
```
OrangeTux, line 129: Does it make sense to create here a `ChargePoint` instance?

OrangeTux:
Thanks for your effort! By using type hints and creating interfaces you made this PR easy to read. Also the examples are helpful to express your ideas. IMHO this PR is much better in transferring your ideas than  work in https://github.com/villekr/ocpp-serverless-example. 

I'm still trying to understand everything, but I can already share some first thoughts. 

I'm not sure if I like the stateless example. For two reasons. First the handler is not stateless as it receives a reference to the `ChargePoint` instance via `self`. Second the signature of the method decorated by `@on()` changes completely. The `@on()` decorator can now be used on 2 total different types. I think that would be very confusing for users of this library. Support for stateless routing should be more explicit. 

I'm also not sure about adding an optional `Context` attribute to all several of the method on `Router` like `route_message()` and `call()`. I'm not sure if we should 'pollute' the API that works for most use case  with arguments to support a rather specific use case. Maybe this can be avoided by creating a subclass of `Router` and add these modification. While leaving the current implementation (nearly) untouched. 

Although it sometimes takes some time before I respond I really appreciate your effort!

villekr:
Thanks for checking out and giving feedback for this work in progress version. For stateless routing I think it's unavoidable that @on() and @after() signatures have to be changed as compared to current ones and that's of course unfortunate. In stateless example ChargePoint instance is not involved but Router base class is, which has still few member attributes defined. So in that sense yes, it's not as stateless as I'd like it to be. 

This draft manages to maintain backwards compatibility both for ASGI stateful and ChargePoint usage. However, it hangs on how ChargePoint is implemented. So in many ways it would propably lead to more understandable interface to just break backwards compatibility when using ASGI interface and keep ChargePoint usage as is.

PR #172, opened 2021-01-26
Add dataclasses for ChargingProfile and it's constituents

grutts:
Some questions:

* These objects don't quite fit into calls, because they're not call payloads. On the other hand, they don't really fit in Enums either. I considered adding a new file like `types.py`, but figured it best to see what your preference is?

* I also wanted to add a check to ensure a SetChargingProfilePayload is serialised to the correct representation when using these objects. I noticed the codebase doesn't have such tests for other payloads, so I left it. Would you like to keep it this way?

Thanks!

### DIFF
--- a/ocpp/v16/call.py
+++ b/ocpp/v16/call.py
@@ -1,3 +1,4 @@
+from decimal import Decimal
 from typing import Any, Dict, List, Optional
 from dataclasses import dataclass, field
 
@@ -5,12 +6,14 @@ from ocpp.v16.enums import (
     AvailabilityType,
     ChargePointErrorCode,
     ChargePointStatus,
+    ChargingProfileKindType,
     ChargingProfilePurposeType,
     ChargingRateUnitType,
     DiagnosticsStatus,
     FirmwareStatus,
     MessageTrigger,
     Reason,
+    RecurrencyKind,
     ResetType,
     UpdateType,
 )
@@ -123,9 +126,44 @@ class SendLocalListPayload:
 
 
 @dataclass
+class ChargingSchedulePeriod:
+    """Charging schedule period structure defines a time period in a charging schedule, as used in: ChargingSchedule."""
+    start_period: int
+    limit: Decimal
+    number_phases: Optional[int] = None
+
+
+@dataclass
+class ChargingSchedule:
+    """Charging schedule structure defines a list of charging periods, as used in: GetCompositeSchedule.conf and
+    ChargingProfile."""
+
+    charging_rate_unit: ChargingRateUnitType
+    charging_schedule_period: List[ChargingSchedulePeriod]
+    duration: Optional[int] = None
+    start_schedule: Optional[str] = None
+    min_charging_rate: Optional[Decimal] = None
+
+
+@dataclass
+class ChargingProfile:
+    """A ChargingProfile consists of a ChargingSchedule, describing the amount of power or current that can be
+    delivered per time interval."""
+    charging_profile_id: int
+    stack_level: int
+    charging_profile_purpose: ChargingProfilePurposeType
+    charging_profile_kind: ChargingProfileKindType
+    charging_schedule: ChargingSchedule
+    transaction_id: Optional[int] = None
+    recurrency_kind: Optional[RecurrencyKind] = None
+    valid_from: Optional[str] = None
+    valid_to: Optional[str] = None
+
+
+@dataclass
 class SetChargingProfilePayload:
     connector_id: int
-    cs_charging_profiles: Dict
+    cs_charging_profiles: ChargingProfile
 
 
 @dataclass


ocpp/v16/call.py
```diff
@@ -122,10 +125,46 @@ class SendLocalListPayload:
     local_authorization_list: List = field(default_factory=list)
 
 
+@dataclass
+class ChargingSchedulePeriod:
+    """Charging schedule period structure defines a time period
+     in a charging schedule, as used in: ChargingSchedule."""
+    start_period: int
+    limit: Decimal
```
OrangeTux, line 133: `Decimal` aren't JSON-serializable. Can you use regular floats here?

OrangeTux (changes requested):
Thanks for your PR! 

I'm unsure what to do this PR. I think your code is an improvement and should be included at some point.. However, the changes are backwards incompatible with the current code. 

There are more backwards incompatible changes that we want to integrate on our way to a v1.0.0 release. I'll discuss with others how and what to do.

In the mean time there are some things that you can improve to your code.

Nested dataclasses can't be deserialized. Consider the following example:


``` python
from dataclasses import asdict
from ocpp.v16.call import ChargingSchedule, ChargingProfile, ChargingSchedulePeriod
from ocpp.v16.enums import ChargingProfilePurposeType, ChargingProfileKindType, ChargingRateUnitType

charging_schedule_period = { 'start_period': 0, 'limit': 10 }

charging_schedule = {
        'charging_rate_unit': ChargingRateUnitType.amps,
        'charging_schedule_period': [charging_schedule_period],
    }

charging_profile = {
    'charging_profile_id': 1,
    'stack_level': 0,
    'charging_profile_purpose': ChargingProfilePurposeType.tx_profile,
    'charging_profile_kind': ChargingProfileKindType.absolute,
    'charging_schedule': charging_schedule,
}


profile = ChargingProfile(**charging_profile)

# This fails as `profile.charging_schedule` is of type `dict`.
assert type(profile.charging_schedule) == ChargingSchedule
```

Can you you modify the code and make sure that nested dataclasses are deserialized correctly? I think you can use [`__post_init__()`](https://docs.python.org/3/library/dataclasses.html?highlight=asdict#post-init-processing) for that.

Regarding the location, I suggest to put the code in it's own module `ocpp.v16.charging_profile`. It doesn't really fit in `ocpp.v16.call` as that module contains only payloads to construct OCPP Calls. It also doesn't fit in `ocpp.v16.enums` because it isn't an enum.

ocpp/v16/call.py
grutts: Yup no problem I'll change that. I had thought, perhaps mistakenly, that as `ocpp.messages._DecimalEncoder` is used in the serialisation of `Call` it would be serialised correctly.

grutts:
No problem!

Good point regarding the de-serialisation, I had missed that. I'll look to make changes to support this.

Apologies, I don't quite understand the backwards incompatibility. My understanding when the payload is passed to `Call` all dataclasses are converted to dictionaries (similarly before any validation), and there is no enforcing of the types on the dataclass, and thus any code which is passing dictionaries to cs_charging_profiles would continue to work. Similarly, my understanding is we don't de-serialise back to the payload class automatically anywhere (nor validate against the dataclass). I do see a backwards incompatibility in static type checking (i.e. mypy), and if this is your concern I would propose we modify the type to still accept dicts (`cs_charging_profiles: Untion[ChargingProfile, Dict]`) ? Or am I missing something?

ocpp/v16/call.py
OrangeTux: You have a fair point. I forgot about that. However, I still suggest to change to floats. I don't see any big advantages to use `Decimal` of `float`. In that case I prefer `float` as this type is way more used. Or do you have any strong argument to use `Decimal`?

OrangeTux:
Changing type of `SetChargingProfilePayload.cs_charging_profiles` from `Dict` to `ChargingProfile` is a backwards incompatible change.

However, you're right when you write "my understanding is we don't de-serialise back to the payload class automatically anywhere ". In practice it means users implementing a `@on()` handler for `SetChargingProfile` still get `cs_charging_profiles` as a `Dict`.

Changing the type hints, as you suggested, is required to express that `cs_charging_profiles` can be both `Dict` or `ChargingProfile`. And with this modification this PR is backwards compatible again.

I think for v1.0.0 we should drop `Dict` as possible type for `cs_charging_profiles` and only allow `ChargingProfile`. That means this library should automatically deserialize incoming `SetChargingProfile` requests into `ChargingProfile` and the other nested dataclasses. But these change are in the scope of a different issue.

grutts:
Ok, I follow you now. I'll go with Union for now then to avoid making a backwards incompatible change.

I agree with the changes on (de)serialisation for v1.0.0 - we're doing this internally on top of the library at the minute at it would be handy to have it upstream!

grutts:
remove_nones needed to change to remove keys where the value is None for nested dicts, and dicts which contains lists of dicts. These structures appear in the charging profile code. I've added that in this commit https://github.com/mobilityhouse/ocpp/pull/172/commits/19bfe11a87e35cfa296a9d2ee68c33e834de975c

If we agree that deserialisation of dicts to dataclasses is out of scope and something for v1.0 then I think this PR is ready to go.

proelke:
There is a PR I'd like to submit and was working on a variation of your change to remove_nones. +1 on that change from me.

ocpp/v16/call.py
tropxy: @grutts did you have time to check this?

grutts: @tropxy ah I forgot to address this. I've changed the types to float in the latest commit.

I also moved the dataclasses to a new `charging_profile` module and added `ChargingProfile` as an acceptable type to the `RemoteStartTransaction` payload (alongside `Dict` and `None`)

grutts:
@OrangeTux @tropxy I think this is ready to go. Are there any other changes you need?

jerome-benoit:
The PR is quite interesting in the context of smart charging algos conception and validation, which has become a requirement for charging infrastructure. 

What is the rationale for next merging it since 3 years?

PR #208, opened 2021-04-30
handle exception in route_message

maurerle:
I don't know if it is better to handle this in the loop or outside the loop but currently I have the behavior that Schema violations are not logged.

This was seen using the html file from https://github.com/JavaIsJavaScript/OCPP-2.0-CP-Simulator.
In this project the modem iccid contains more than 20 chars like this:  "iccid": "MMCC IINN NNNN NNNN NN C x",

Using the proposed fix the Logger gives feedback to the user. Otherwise the connection is closed without a reason and the websocket connection returns with 1011

### DIFF
--- a/examples/v201/central_system.py
+++ b/examples/v201/central_system.py
@@ -65,8 +65,10 @@ async def on_connect(websocket, path):
 
     charge_point_id = path.strip('/')
     charge_point = ChargePoint(charge_point_id, websocket)
-
-    await charge_point.start()
+    try:
+        await charge_point.start()
+    except Exception as e:
+        logging.exception(e)
 
 
 async def main():
--- a/ocpp/charge_point.py
+++ b/ocpp/charge_point.py
@@ -123,7 +123,11 @@ class ChargePoint:
             message = await self._connection.recv()
             LOGGER.info('%s: receive message %s', self.id, message)
 
-            await self.route_message(message)
+            try:
+                await self.route_message(message)
+            except Exception as e:
+                LOGGER.exception(e)
+
 
     async def route_message(self, raw_msg):
         """
@@ -163,10 +167,11 @@ class ChargePoint:
         except KeyError:
             raise NotImplementedError(f"No handler for '{msg.action}' "
                                       "registered.")
-
-        if not handlers.get('_skip_schema_validation', False):
-            validate_payload(msg, self._ocpp_version)
-
+        try:
+            if not handlers.get('_skip_schema_validation', False):
+                validate_payload(msg, self._ocpp_version)
+        except Exception as e:
+            LOGGER.exception(e)
         # OCPP uses camelCase for the keys in the payload. It's more pythonic
         # to use snake_case for keyword arguments. Therefore the keys must be
         # 'translated'. Some examples:
--- a/ocpp/charge_point.py
+++ b/ocpp/charge_point.py
@@ -167,11 +167,10 @@ class ChargePoint:
         except KeyError:
             raise NotImplementedError(f"No handler for '{msg.action}' "
                                       "registered.")
-        try:
-            if not handlers.get('_skip_schema_validation', False):
-                validate_payload(msg, self._ocpp_version)
-        except Exception as e:
-            LOGGER.exception(e)
+
+        if not handlers.get('_skip_schema_validation', False):
+            validate_payload(msg, self._ocpp_version)
+
         # OCPP uses camelCase for the keys in the payload. It's more pythonic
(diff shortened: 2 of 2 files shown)


maurerle:
I think its enough to change the examples for this, so that other people don't run into this kind of errors

examples/v201/central_system.py
```diff
@@ -65,8 +65,10 @@ async def on_connect(websocket, path):
 
     charge_point_id = path.strip('/')
     charge_point = ChargePoint(charge_point_id, websocket)
-
-    await charge_point.start()
+    try:
+        await charge_point.start()
+    except Exception as e:
+        logging.exception(e)
```
laysauchoa, line 71: logging.exception already contains the stack trace alongside.
```
In [1]: def start(): 
   ...:     raise KeyError 


In [6]: try: 
   ...:     start() 
   ...: except Exception: 
   ...:     logging.exception("Something happened") 
   ...:                                                                                             
ERROR:root:Something happened
Traceback (most recent call last):
  File "<ipython-input-6-edb1dc331440>", line 2, in <module>
    start()
  File "<ipython-input-1-190264bc2ae3>", line 2, in start
    raise KeyError
KeyError
```
So you can use that and add some custom message about it ;)

OrangeTux:
Thanks for the PR. I'm not really seeing the value of this change. The goal of this example is to demonstrate how to create a very simple CSMS. It's goal is not to be production ready app. IMHO error handling is only distracting from what that goal.

maurerle:
I disagree. This is a small working example from which people will copy and derive their implementation.
It took me quite some time to get a clue that there are error messages which are swallowed in the thread.

The current showcase doesnt print the error if someone wants to add a command but has a typo (or the wrong parameter names).
It just terminates for whatever reason, without feedback.

Its not becoming production ready just because you give the user feedback on what he did wrong, does it?

OrangeTux:
> The current showcase doesnt print the error if someone wants to add a command but has a typo (or the wrong parameter names). It just terminates for whatever reason, without feedback.

I'm not sure what you mean. I introduced a typo in this patch:

``` diff
diff --git a/examples/v201/central_system.py b/examples/v201/central_system.py
index f76217f..ac106c5 100644
--- a/examples/v201/central_system.py
+++ b/examples/v201/central_system.py
@@ -31,7 +31,7 @@ class ChargePoint(cp):
     @on('Heartbeat')
     def on_heartbeat(self):
         print('Got a Heartbeat!')
-        return call_result.HeartbeatPayload(
+        return call_result.eartbeatPayload(
             current_time=datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%S') + "Z"
         )
```

When I connect a client to this CSMS the error is shown without calling `logging.exception()`.

```
INFO:root:Server Started listening to new connections...
INFO:root:Protocols Matched: ocpp2.0.1
INFO:ocpp:CP_1: receive message [2,"bd6e6984-3be2-42dc-b63a-c22e872d124d","BootNotification",{"chargingStation":{"model":"Wallbox XYZ","vendorName":"anewone"},"reason":"PowerUp"}]
INFO:ocpp:CP_1: send [3,"bd6e6984-3be2-42dc-b63a-c22e872d124d",{"currentTime":"2021-06-29T19:26:54.561113","interval":10,"status":"Accepted"}]
INFO:ocpp:CP_1: receive message [2,"11bbadae-9a2f-40dc-ab62-9f84b1073978","Heartbeat",{}]
Got a Heartbeat!
ERROR:ocpp:Error while handling request '<Call - unique_id=11bbadae-9a2f-40dc-ab62-9f84b1073978, action=Heartbeat, payload={}>'
Traceback (most recent call last):
  File "/home/auke/projects/ocpp/ocpp/charge_point.py", line 185, in _handle_call
  File "/home/auke/projects/ocpp/ocpp/routing.py", line 47, in inner
    return func(*args, **kwargs)
  File "/home/auke/projects/ocpp/examples/v201/central_system.py", line 34, in on_heartbeat
    return call_result.eartbeatPayload(
AttributeError: module 'ocpp.v201.call_result' has no attribute 'eartbeatPayload'
INFO:ocpp:CP_1: send [4,"11bbadae-9a2f-40dc-ab62-9f84b1073978","InternalError","An unexpected error occurred.",{}]
```

maurerle:
Yes, when handling a request, the error is logged to ocpp

If the message from the client is not according to the specification, the error does not get logged.
Same if the handler name is wrong (i.e. @on('eartbeat'))

At least thats, what I experienced 3 Months ago. But i can also reproduce an example tomorrow

PR #214, opened 2021-06-13
tests: fix validation of date-time format fields

laysauchoa:
In JSON format_checker=FormatChecker() should be
explicitly passed so the format of the field is
checked as it is None by default.

class jsonschema.Draft4Validator(schema, types=(),
resolver=None, format_checker=None)

For more info:
https://python-jsonschema.readthedocs.io/en/stable/validate/

### DIFF
--- a/tests/test_messages.py
+++ b/tests/test_messages.py
@@ -156,7 +156,7 @@ def test_validate_payload_with_valid_payload(ocpp_version):
     message = CallResult(
         unique_id="1234",
         action="Heartbeat",
-        payload={'currentTime': datetime.now().isoformat()}
+        payload={'currentTime': datetime.now().isoformat()+'Z'}
     )
 
     validate_payload(message, ocpp_version=ocpp_version)
--- a/ocpp/messages.py
+++ b/ocpp/messages.py
@@ -184,7 +184,8 @@ def get_validator(
     #     Unexpected UTF-8 BOM (decode using utf-8-sig):
     with open(path, 'r', encoding='utf-8-sig') as f:
         data = f.read()
-        validator = Draft4Validator(json.loads(data, parse_float=parse_float), format_checker=FormatChecker())
+        validator = Draft4Validator(json.loads(data, parse_float=parse_float),
+                                    format_checker=FormatChecker())
         _validators[cache_key] = validator
 
     return _validators[cache_key]
--- a/ocpp/messages.py
+++ b/ocpp/messages.py
@@ -7,7 +7,7 @@ import warnings
 from typing import Callable, Dict
 from dataclasses import asdict, is_dataclass
 
-from jsonschema import Draft4Validator
+from jsonschema import Draft4Validator, FormatChecker
 from jsonschema.exceptions import ValidationError as SchemaValidationError
 
 from ocpp.exceptions import (OCPPError, FormatViolationError,
@@ -184,7 +184,7 @@ def get_validator(
     #     Unexpected UTF-8 BOM (decode using utf-8-sig):
     with open(path, 'r', encoding='utf-8-sig') as f:
         data = f.read()
-        validator = Draft4Validator(json.loads(data, parse_float=parse_float))
+        validator = Draft4Validator(json.loads(data, parse_float=parse_float), format_checker=FormatChecker())
         _validators[cache_key] = validator
 
     return _validators[cache_key]
--- a/tests/v16/test_v16_charge_point.py
+++ b/tests/v16/test_v16_charge_point.py
@@ -24,7 +24,7 @@ async def test_route_message_with_existing_route(base_central_system,
         assert kwargs['firmware_version'] == "#1:3.4.0-2990#N:217H;1.0-223"
 
         return call_result.BootNotificationPayload(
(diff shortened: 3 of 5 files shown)


OrangeTux:
The schemas for OCPP haven't define any `format`s. So I don't any effect of the this PR. Therefore I suggest not to merge it. Let me know if you think otherwise.

laysauchoa:
Hi, thanks for reviewing this. 
I think otherwise:

- OCPP specification refers to this `currentTime` as of type `dateTime`. Currently, if this field is replaced by any random string, it would pass validation. As explained in the [documentation](https://python-jsonschema.readthedocs.io/en/stable/validate/):

> format_checker – an instance of FormatChecker whose FormatChecker.conforms method will be called to check and see if instances conform to each format property present in the schema. If unprovided, no validation will be done for format. 

> The schemas for OCPP haven't define any formats

I am not sure if I understand you correct. I see that we have a `format` specified on the schemas, for e.g. `BootNotificationResponse.json`:

```
        "currentTime": {
            "type": "string",
            "format": "date-time"
        }
```


Was this format not intentional and  should it be removed? or could you please clarify  your thoughts? Would another format checker suit more in this case?

OrangeTux:
My bad, I ran `grep format` in the wrong directory. I also missed the #216 that is related to this PR.

But still I'm not entirely sure if we should merge this PR.  [JSON schema's validates the 'date-time' format according to ISO8601](https://json-schema.org/understanding-json-schema/reference/string.html#format). However, the OCPP 1.6 doesn't specify any format for the `datetime` type. I'm not entirely sure if we should assume ISO8601 here.

laysauchoa:
I understand. Maybe the right format can be discussed further, but not picking any format makes any string passes this validation, as there will be no validation in this case.

tests/test_messages.py
```diff
@@ -185,7 +185,7 @@ def test_validate_payload_with_valid_payload(ocpp_version):
     message = CallResult(
         unique_id="1234",
         action="Heartbeat",
-        payload={'currentTime': datetime.now().isoformat()}
+        payload={'currentTime': datetime.now().isoformat()+'Z'}
```
tropxy, line 188: for consistency, I suggest we use the datetime result of this:
`datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%S') + "Z"`
which is what we use in ocpp/routing.py

Also, since it is related, maybe you can fix all the fields where we are incorrectly using the datetime and substitute them by the line above

tropxy (changes requested):

Jared-Newell-Mobility:
@OrangeTux - if this is only for consistency maybe @tropxy can be considered, what are you thoughts?

OrangeTux:
I'm afraid we can't merge this. 

OCPP 2.0.1 is clear about the required format for date time objects:

> All time values exchanged between CSMS and Charging Station SHALL be formatted as
defined in [RFC3339]. Additionally fractional seconds have been given an extra limit. The
number of decimal places SHALL NOT exceed the maximum of 3.

`jsonschemas` validates date-time  [according to ISO8601](https://json-schema.org/understanding-json-schema/reference/string#built-in-formats). 

> Dates and times are represented in [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6). This is a subset of the date format also commonly known as [ISO8601 format](https://www.iso.org/iso-8601-date-and-time-format.html).

In other words: `jsonschema` is more strict than the OCPP spec allows.

This website makes the difference visual: https://ijmacd.github.io/rfc3339-iso8601/

Jared-Newell-Mobility:
Thanks for the explanation, it gives clear references:

Currently,

datetime.now().isoformat()
'2023-10-20T11:52:42.538068'

Technically, in the tests at least (presented here) violate both:

OCPP 2.0.1 - SHALL NOT exceed the maximum of 3
and
RFC 3339 - by not having a Local Offset or UTC

I haven't reopened this, but interested in your thoughts?

OrangeTux:
>  RFC 3339 - by not having a Local Offset or UTC

Where did you find this requirement? I can't find it in the OCPP 2.0.1 Part 2 - Specification

I think we've to roll our own validation. That should be possible according to the[ json-schema's docs:](https://python-jsonschema.readthedocs.io/en/latest/api/#jsonschema.FormatChecker.checks):

> To add a check for a custom format use the [FormatChecker.checks](https://python-jsonschema.readthedocs.io/en/latest/api/#jsonschema.FormatChecker.checks) decorator.

Jared-Newell-Mobility:
Section 2.1.4 Primitive Datatypes 

"All time values exchanged between CSMS and Charging Station SHALL be formatted as
defined in [RFC3339]. Additionally fractional seconds have been given an extra limit. The
number of decimal places SHALL NOT exceed the maximum of 3.
Example 1: 2019-04-12T23:20:50.52Z represents 20 minutes and 50.52 seconds after the 23rd
hour of April 12th, 2019 in UTC.
Example 2: 2019-12-19T16:39:57+01:00 represents 39 minutes and 57 seconds after the 16th
hour of December 19th, 2019 with an offset of +01:00 from UTC (Central European Time)"

You are right that our own validation would be the best solution

PR #215, opened 2021-06-13
improvement: use isinstance instead of type

laysauchoa:
`isinstance()` accounts for inheritance, while
`type()` quality check does not.

for more info: https://stackoverflow.com/questions/1549801/what-are-the-differences-between-type-and-isinstance

### DIFF
--- a/ocpp/messages.py
+++ b/ocpp/messages.py
@@ -214,10 +214,10 @@ def validate_payload(message, ocpp_version):
         # Both the schema and the payload must be parsed using the different
         # parser for floats.
         if ocpp_version == '1.6' and (
-            (type(message) == Call and
+            (isinstance(message, Call) and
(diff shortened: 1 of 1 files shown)


OrangeTux:
Thanks for the PR. I don't see why `type()` is an improvement over `isinstance()` in this case. But I do see a drawback: the behavior of `validate_payload()` might change for some users. So therefore I'm hesitating to merge this PR. Can you elaborate on why `type()` is preferred over `instance()` in this situation?

laysauchoa:
Hey, 
Thanks for reviewing. This is more like a general best practice to use it because this can happen:

```
class MyList(list):
    pass
   
>>> my_list = MyList()
>>> type(my_list) is list
False
>>> isinstance(my_list, list)
True
```

In this case, we are comparing it correct and we would not face this issue, so I understand your point and I agree that does not make sense to merge it.

Issue #223, opened 2021-06-18
DataTransferPayload status validation


drc38:
Should this `status` in `call_result.py` be updated to `status: DataTransferStatus` rather than a generic string?

@dataclassclass 
DataTransferPayload:    
status: str    
data: Optional[str] = None

laysauchoa:
@drc38 I agree with you. https://github.com/mobilityhouse/ocpp/pull/224 closes #223

OrangeTux:
Thanks for reporting!

PR #224, opened 2021-06-18
fix: use DataTransferStatus in status

laysauchoa:
In ocpp 1.6, the field status is of type
DataTransferStatus, as we have it defined,
this is what we should be using for the
status field of DataTransferPayload.


OrangeTux (approved):

OrangeTux:
Fixes #223

PR #231, opened 2021-07-06
add ConfigurationKeys

lbbrhzn:
added configuration keys (for v16 only)


lbbrhzn:
i added the configuration keys to enums.py, not sure if you need to update your schema_to_enums.pu script to prevent them from being overwritten.

lbbrhzn:
Also, might be nice to separate the keys for the different profiles. What would be a good way to do that?

tropxy:
Thanks for this @lbbrhzn, I guess you manually added all the keys, right? in order to avoid losing data, what I think we could do is to add an extra file, containing the original OCPP config keys, for 1.6 and read from there when we run the schema_to_enums.py, but not sure if it is the best idea

lbbrhzn:
Not sure either. Perhaps that's something you can do in a new pull request?

OrangeTux:
This PR is superseded by #413. Thanks @lbbrhzn

PR #249, opened 2021-09-15
Remove depreciated get_schema code

proelke:
When running tests I noticed the following warning: `UserWarning: Depricated as of 0.8.1. Please use `ocpp.messages.get_validator()`

As the next revision will be 0.8.4, I figured I would submit a PR to remove this warning.

### DIFF
--- a/tests/test_messages.py
+++ b/tests/test_messages.py
@@ -79,6 +79,14 @@ def test_get_validator_with_valid_name():
     }
 
 
+def test_get_validator_with_invalid_name():
+    """
+    Test if OSError is raised when schema validation file cannnot be found.
+    """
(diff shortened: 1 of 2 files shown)


tests/test_messages.py
```diff
@@ -168,14 +140,6 @@ def test_validate_get_composite_profile_payload():
     validate_payload(message, ocpp_version="1.6")
 
 
-def test_get_schema_with_invalid_name():
```
tropxy, line 171: do we have a test_get_validator_with_invalid_name ?

proelke: Nope but there is now.

OrangeTux (approved):
Thanks!

PR #250, opened 2021-09-16
Add v1.6 data types

proelke:


ocpp/v16/datatypes.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+@dataclass
+class AuthorizationData:
+    """
+    Elements that constitute an entry of a Local Authorization List update.
+    """
+
+    id_tag: IdToken
+    id_tag_info: IdTagInfo
```
OrangeTux, line 51: This field is optional.

ocpp/v16/datatypes.py
```diff
@@ -0,0 +1,184 @@
...
+@dataclass
+class IdTagInfo:
+    """
+    Contains status information about an identifier. It is returned in
+    Authorize, Start Transaction and Stop Transaction responses.
+
+    If expiryDate is not given, the status has no end date.
+    """
+
+    status: AvailabilityStatus
```
OrangeTux, line 39: This type is AuthorizationStatus

OrangeTux (changes requested):
I suggest to remove the `CiStringXXXType` and the `IdToken` All other classes model a type; a bag of attributes. The `CiStringXXXType` and `IdToken` are merely a string with some limits on their length. 

It also creates a non-friendly API. To read the value of the an `IdTagInfo`:

``` python
assert IdTagInfo.parent_id_tag.id_token.ci_string_20 == 'my-rfid'
```

Ideally, I want the API to be:

``` python
assert IdTagInfo.parent_id_tag == 'my-rfid'
```

I'm not sure yet, a nice way to impose this length on attributes of type `CiStringXXXType` and `IdToken`. So for now I suggest that the length is not validated and replace the type hints of attributes using those types with `str`. What do yo think?

proelke:
I was debating adding them from the beginning. Then I debated adding some kind of custom validator, but I figured you might be against the complexity for such a small thing. To avoid a dependency I could use a __post_init__ to verify the length of the string in IdTagInfo and remove the CiStringXXXType. Or I can just drop it all together and leave it as is.

ocpp/v16/datatypes.py
proelke: Fixed.

ocpp/v16/datatypes.py
proelke: Fixed.

OrangeTux:
> To avoid a dependency I could use a post_init to verify the length of the string in IdTagInfo and remove the CiStringXXXType

I like that solution

OrangeTux (changes requested):
The code looks good. Make sure to fix the lint errors so the CI job passes.

proelke:
@OrangeTux fixed up the linting issues

PR #256, opened 2021-10-05
Add support for Python 3.10

laysauchoa:
Python 3.10 has been released on 4 Oct, 2021.
We can add support for it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,6 +43,18 @@ jobs:
           password: $DOCKERHUB_PASSWORD
     <<: *steps
 
+  test-310:
+    docker:
+      - image: circleci/python:3.10
+        auth:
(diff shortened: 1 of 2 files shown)


laysauchoa:
oh, this image is not yet available.I'll leave it here :)

OrangeTux:
The image is available now. However, the build fails.

tropxy:
There are some incompatibilities with python 3.10 and older versions of pytest [1]. The problem may be there...

[1]: https://stackoverflow.com/questions/69528842/an-error-while-trying-to-execute-tests-on-python-3-10-with-pytest

tropxy:
As we moved to GHA and 3.10 tests were included there, this PR can be closed. Thanks for the initiative @laysauchoa

PR #259, opened 2021-10-21
Update enums.py

shaikhasif15752:
Fixed SetMonitorBase typo, which was causing NotImplementedError


tropxy:
There's some weird error with the CI tests when import the six package.. @OrangeTux can we re-run the build? I checked if I could and I cant...

OrangeTux (approved):
Thanks for the PR.

Issue #252, opened 2021-09-21
Tests started to fail in CircleCI


laysauchoa:
Locally, I have run:
```
make deps && make test
```
But this step is failing in the circleci [here](https://app.circleci.com/pipelines/github/mobilityhouse/ocpp/427/workflows/c4d6a24e-cc7a-4d9c-b2ee-1278b03b11b2/jobs/1263) and basically just a text on [README](https://github.com/mobilityhouse/ocpp/commit/a454235df30f9dce405a49219f4f3e7cd239a1de) was changed as a possible test.

OrangeTux:
Interesting catch. I [reran a build](https://app.circleci.com/pipelines/github/mobilityhouse/ocpp?branch=master) that previously worked and this build starts to fail now as well. It seems that the build is not reproducible.

OrangeTux:
It seems that a recent version of `poetry` contains a bug. See https://github.com/python-poetry/poetry/issues/4523  As the build doesn't pins the version of `poetry` the latest version is installed. Thanks @Roger for figuring this out.

Roger:
After digging a bit more, looks like our issue is that our lockfile is in an old format that doesn't contain the algo in the hashes and since we didn't add any dependency to the project for so long, it didn't update to the latest format and at some point became backwards incompatible https://github.com/python-poetry/poetry/issues/4523#issuecomment-925075103

Running `poetry lock --no-update` should fix the issue

laysauchoa:
@Roger this is what I get when I run poetry `lock --no-update`. 

```
✦ ➜ poetry lock --no-update

[NoSuchOptionException]
The "--no-update" option does not exist.

Traceback (most recent call last):
  File "/home/tmhdev/.local/lib/python3.6/site-packages/clikit/console_application.py", line 124, in run
    self, args, input_stream, output_stream, error_stream
  File "/home/tmhdev/.local/lib/python3.6/site-packages/poetry/console/config/application_config.py", line 165, in create_io
    resolved_command = application.resolve_command(args)
  File "/home/tmhdev/.local/lib/python3.6/site-packages/clikit/console_application.py", line 110, in resolve_command
    return self._config.command_resolver.resolve(args, self)
  File "/home/tmhdev/.local/lib/python3.6/site-packages/clikit/resolver/default_resolver.py", line 34, in resolve
    return self.create_resolved_command(result)
[...]

PR #262, opened 2021-10-28
Fix CI build 

OrangeTux:
`poetry` introduced some [breaking changes] in the format of the lock
file[0]. This project didn't pin the version of `poetry`, therefore the
build started to fail.

This has been fixed by pinning `poetry` and updating the lock file.

Fixes: #252

[breaking changes]: python-poetry/poetry#4523

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,15 @@
 version: 2.1
+
+steps: &steps
+  steps:
+    - checkout
+    - run:
+          name: run tests
+          command: |
+              # Poetry isn't installed in the image and need to be installed manually.
+              sudo pip install poetry==1.1.11
+              make deps && make test
+
 jobs:
   test-36:
     docker:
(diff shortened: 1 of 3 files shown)


tropxy:
> `poetry` introduced some [breaking changes](python-poetry/poetry#4523) in the format of the lock file[0]. This project didn't pin the version of `poetry`, therefore the build started to fail.
> 
> This has been fixed by pinning `poetry` and updating the lock file.
> 
> Fixes: #252

the breaking changes link is not working

.circleci/config.yml
```diff
@@ -12,7 +12,7 @@ jobs:
             name: run tests
             command: |
                 # Poetry isn't installed in the image and need to be installed manually.
-                sudo pip install poetry
+                sudo pip install poetry==1.1.11
```
tropxy, line 15: could you add the poetry version as an env on top that we can change easily, so that we dont have to type the version explicitly all the time?

tropxy (changes requested):
check it out

.circleci/config.yml
OrangeTux: Make sense. I applied the requested changes.

tropxy (approved):
sweet

PR #268, opened 2021-11-21
Move to GHA and Makefile improvements

### LINKED ISSUE #245: Automate release to Pypi.
Currently releases are done manually . So I have to:

* update CHANGELOG.md
* update version numbers in pyproject.toml and docs/source/conf.py
* commit changes to master
* tag commit with correct release
* push commit and tag to Github.
* build the packages using `poetry build`
* upload the packages to pypi.org using `twine upload dist/*`

This process should be automate to increase velocity of releases and to make release less dependent on me.

tropxy:
fixes #245

* Improved Makefile to include an helper and checks before installing dependencies;
* Added GitHib actions to automatically run the tests and publish the repo based on semantic versioning.

### DIFF
--- a/.github/workflows/publish-to-pypi.yml
+++ b/.github/workflows/publish-to-pypi.yml
@@ -1,0 +1,34 @@
+name: Publish Python 🐍 distributions 📦 to Switch PyPI
+
+on:
+  push:
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+*"
+
+jobs:
+  build-n-publish:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - name: Set up Python 3.10.0
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.10.0
+
+    - name: Gets the tag version
+      id: context
+      run: |
+        echo ::set-output name=TAG_VERSION::${GITHUB_REF#refs/tags/}
+
+    - name: Setup the Python Environment by installing Poetry
+      uses: ./.github/actions/setup-python-build-env
+
+    - name: Poetry bump version, build and publish
+      shell: bash
+      run: |
+        poetry version $TAG_VERSION
+        poetry update
+        poetry publish --build -u __token__ -p $PYPI_TOKEN
+      env:
+        TAG_VERSION: ${{ steps.context.outputs.TAG_VERSION }}
+        PYPI_TOKEN: ${{ secrets.PYPI_API_TOKEN }}
--- a/Makefile
+++ b/Makefile
@@ -1,20 +1,45 @@
-.PHONY: deps docs test build deploy
+.PHONY: help .install-poetry update install docs tests build deploy
+
+.DEFAULT_GOAL := help
+
+IS_POETRY := $(shell pip freeze | grep "poetry==")
+IS_TWINE := $(shell pip freeze | grep "twine==")
+
+
+help:
+	@echo "Please use 'make <target>' where <target> is one of"
+	@echo ""
+	@echo "  help           outputs this helper"
+	@echo "  build          builds the project .whl with poetry"
+	@echo "  update         updates the dependencies in poetry.lock"
+	@echo "  install        installs the dependencies in the env"
+	@echo "  test           run all the tests and linting"
+	@echo "  deploy         deploys the project using twine (not recommended, only use if really needed)"
+	@echo ""
+	@echo "Check the Makefile to know exactly what each target is doing."
+
 
 export PATH := ${HOME}/.local/bin:$(PATH)
 
-deps:
-	pip install poetry
+.install-poetry:
+	@if [ -z ${IS_POETRY} ]; then pip install poetry; fi
+
+update: .install-poetry
+	poetry update
+
+install: .install-poetry
 	poetry install
 
-docs:
+docs: .install-poetry
 	poetry run sphinx-build -b html docs/source docs/build
 
-test:
+tests: .install-poetry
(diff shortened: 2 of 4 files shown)


.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,34 @@
...
+      run: |
+        echo ::set-output name=TAG_VERSION::${GITHUB_REF#refs/tags/}
+
+    - name: Setup the Python Environment by installing Poetry
+      uses: ./.github/actions/setup-python-build-env
+
+    - name: Poetry bump version, build and publish
+      shell: bash
+      run: |
+        poetry version $TAG_VERSION
```
OrangeTux, line 29: TIL about `poetry version`.

Makefile
```diff
@@ -1,20 +1,45 @@
...
+
+help:
+	@echo "Please use 'make <target>' where <target> is one of"
+	@echo ""
+	@echo "  help           outputs this helper"
+	@echo "  build          builds the project .whl with poetry"
+	@echo "  update         updates the dependencies in poetry.lock"
+	@echo "  install        installs the dependencies in the env"
+	@echo "  test           run all the tests and linting"
+	@echo "  deploy         deploys the project using twine (not recommended, only use if really needed)"
```
OrangeTux, line 17: Good idea to add a `help` target. Could you order the targets alphabetically to make it a bit more user friendly?

OrangeTux (changes requested):
Looks good!

.github/workflows/publish-to-pypi.yml
tropxy: Well, I think it is easier if the version comes from the released tag, however I see the gap that this causes in respect to the pyproject.toml. Ideally, the CI job would release the tagged version and update the pyproject.toml, but I couldnt find a way to do this... What we can possibly do is to check if the tagged version matches the version in pyproject.toml and if it does not, aborts the upload, what do you think?

>  Note that update CHANGELOG.md and updating __version__ in doc/source/conf.py is also part of a release.
That is right...not sure how to proceed refarding the CHANGELOG.md, is there any check we can do? This is one of the pieces, that I guess, have to be done manually, right?

OrangeTux: > What we can possibly do is to check if the tagged version matches the version in pyproject.toml and if it does not, aborts the upload, what do you think?

I like that solution. You could run `poetry version <version>` and check if the repo is dirty afterwards. If not, continue to upload the project. If the repo is dirty, abort.

.github/workflows/publish-to-pypi.yml
```diff
@@ -0,0 +1,34 @@
+name: Publish 📦 to PyPi 🚀
+
+on:
+  push:
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+*"
```
OrangeTux, line 6: I just noticed that this project doesn't prefix the tags with `v`. For sake of consistency, i suggest remove the `v`

OrangeTux (changes requested):

.github/workflows/publish-to-pypi.yml
tropxy: check my changes, please

.github/workflows/publish-to-pypi.yml
tropxy: true, will do so

.github/workflows/publish-to-pypi.yml
tropxy: During my edition before, my answer/question got caught by the quote `>`

> Note that update CHANGELOG.md and updating version in doc/source/conf.py is also part of a release.

That is right...not sure how to proceed refarding the CHANGELOG.md, is there any check we can do? This is one of the pieces, that I guess, have to be done manually, right?

OrangeTux: > .not sure how to proceed regarding the CHANGELOG.md, is there any check we can do?

I think there are tools around to generate a CHANGELOG. But that seem a bit overkill IMHO. Maybe you can print a reminder when one runs `make release` saying: 'Don't forget to update CHANGELOG.md'.

tropxy: all right. We can try to cover that in another PR as I was thinking on adding also isort and black for formatting the code, so we could add that together

OrangeTux: > adding also isort and black

Yes, please!

OrangeTux (dismissed):

.github/workflows/publish-to-pypi.yml
tropxy: Check my code, please. I was able to add the CURRENT_VERSION as a shell variable into the target but I couldnt find a way then to expand it into the sed command, so I had to add the variable at the top of the Makefile

Makefile
```diff
@@ -40,8 +43,10 @@ tests: .install-poetry
 build: .install-poetry
 	poetry build
 
-release:
-	poetry version ${version}
+release: .install-poetry
+	@echo "Please remember to update the CHANGELOG.md, before creating the release tag in GitHub"
```
OrangeTux, line 47: I think a more correct wording would be '... before tagging the release'. You don't need to use Github for tagging a commit (at least, I never do that. I just run `git tag <version>`.

OrangeTux (changes requested):
Good addition.

Makefile
tropxy: done

OrangeTux (approved):

PR #278, opened 2021-12-05
Use IdTagInfo datatype instead of Dict for id_tag_info occurances

chan-vince:
The `Authorize.conf` message in the v1.6 spec defines what fields should be present for an `IdTagInfo`. The dataclass for this already exists in `datatypes.py`, as of the recently merged #250 PR.

This PR makes use of it by importing the dataclass, and using it instead of the more losely typed `typing.Dict`.


OrangeTux (approved):
Thanks!

chan-vince:
> Thanks!

Thanks for taking a look. There's actually quite a few similar occurances in dataclasses where things like `typing.List` and `typing.Dict` are used instead of the new specific dataclasses.

I'll have a PR to update those soon..

dinko-pehar:
What happend to this ? I have it like this:
```
call_result.AuthorizePayload(
            id_tag_info={}
        )
```
but instead should be something like:
call_result.AuthorizePayload(
            id_tag_info=IdTagInfo(...)
        )
```
Should the type be changed to `IdTagInfo` ?

diorcety:
Seems pretty correct PR, what is blocking it?

OrangeTux:
Nothing blocks it from being merged. Thanks for the notification @diorcety

Issue #289, opened 2022-01-14
GetConfiguration object with key set to list of config keys results in AttributeError for remove_nones/ChargePoint.call


shirley-li-powerflex:
OCPP library version: 0.12.0
OCPP Version: 1.6J
Python Version: 3.9.7

I was running my tests in a docker container and noticed that whenever I sent a GetConfiguration request, it would result in an AttributeError while converting a message payload to a call object. This actually wasn't an issue until a day ago. I haven't made any changes in my tests for this particular message. Is this expected or am I doing something wrong? I checked the OCPP 1.6 documentation and the GetConfiguration content in the payload seems to be valid. 

```
$ python3 -m asyncio
asyncio REPL 3.9.7 (default, Oct  1 2021, 23:37:25) 
[Clang 12.0.0 (clang-1200.0.32.29)] on darwin
Use "await" directly instead of "asyncio.run()".
Type "help", "copyright", "credits" or "license" for more information.
>>> import asyncio
>>> from ocpp.charge_point import ChargePoint
>>> cp = ChargePoint("cb_id", None)
>>> from ocpp.v16.call import GetConfigurationPayload
>>> get_config_payload  = GetConfigurationPayload(key=["ClockAlignedDataInterval", "ConnectionTimeOut"])
>>> await cp.call(get_config_payload)
Traceback (most recent call last):
  File "/Users/shirleyli/.pyenv/versions/3.9.7/lib/python3.9/concurrent/futures/_base.py", line 445, in result
    return self.__get_result()
  File "/Users/shirleyli/.pyenv/versions/3.9.7/lib/python3.9/concurrent/futures/_base.py", line 390, in __get_result
    raise self._exception
  File "<console>", line 1, in <module>
  File "/Users/shirleyli/.pyenv/versions/3.9.7/lib/python3.9/site-packages/ocpp/charge_point.py", line 263, in call
    payload=remove_nones(camel_case_payload)
  File "/Users/shirleyli/.pyenv/versions/3.9.7/lib/python3.9/site-packages/ocpp/charge_point.py", line 81, in remove_nones
    new_list.append(remove_nones(item))
  File "/Users/shirleyli/.pyenv/versions/3.9.7/lib/python3.9/site-packages/ocpp/charge_point.py", line 75, in remove_nones
    for k, v in dict_to_scan.items():
AttributeError: 'str' object has no attribute 'items'
```

OrangeTux:
[...]

PR #290, opened 2022-01-14
Fix bug in `remove_nones()` when processing `str`

OrangeTux:
`remove_nones()` only works with `list`s and `dict`s, not with other
iterables like `str`. However, the function didn't properly checked its
input and would crash when passed a `list` of `str`s.

Fixes: #289

### DIFF
--- a/ocpp/charge_point.py
+++ b/ocpp/charge_point.py
@@ -4,6 +4,7 @@ import logging
 import uuid
 import re
 import time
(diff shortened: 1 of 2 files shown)


proelke:
Looks good to me. Oops 😬

OrangeTux:
Fixed in [v0.12.1](https://github.com/mobilityhouse/ocpp/tree/0.12.1).

PR #294, opened 2022-01-25
Improved error handling during schema validation

joaomariord:
Added error differentiation when schema is invalid. supporting:
        - <Error in schema> : <Error sent to CP>
        - Type mismatch: TypeConstraintViolationError
        - Lack of properties: ProtocolError
        - Non-requested properties: FormatViolationError

Now catches exceptions thrown by schema validation errors instead of
dropping the WS connection.

Previous behavior was to drop the socket connection whenever a request payload was not compliant with the schema, this PR introduces support for schema error handling according to the OCPP specification (from 1.6 to 2.0.1).

### DIFF
--- a/ocpp/charge_point.py
+++ b/ocpp/charge_point.py
@@ -146,7 +146,12 @@ class ChargePoint:
             return
 
         if msg.message_type_id == MessageType.Call:
-            await self._handle_call(msg)
+            try:
+                await self._handle_call(msg)
+            except OCPPError as error:
+                LOGGER.exception("Error while handling request '%s'", msg)
+                response = msg.create_call_error(error).to_json()
+                await self._send(response)
 
         elif msg.message_type_id in \
                 [MessageType.CallResult, MessageType.CallError]:
@@ -171,7 +176,6 @@ class ChargePoint:
 
         if not handlers.get('_skip_schema_validation', False):
             validate_payload(msg, self._ocpp_version)
-
         # OCPP uses camelCase for the keys in the payload. It's more pythonic
         # to use snake_case for keyword arguments. Therefore the keys must be
         # 'translated'. Some examples:
--- a/ocpp/messages.py
+++ b/ocpp/messages.py
@@ -6,13 +6,13 @@ import decimal
 from typing import Callable, Dict
 from dataclasses import asdict, is_dataclass
 
-from jsonschema import Draft4Validator
+from jsonschema import Draft4Validator, _validators as SchemaValidators
 from jsonschema.exceptions import ValidationError as SchemaValidationError
 
 from ocpp.exceptions import (OCPPError, FormatViolationError,
                              PropertyConstraintViolationError,
-                             ProtocolError, ValidationError,
-                             UnknownCallErrorCodeError)
+                             ProtocolError, TypeConstraintViolationError,
+                             ValidationError, UnknownCallErrorCodeError)
 
 _validators: Dict[str, Draft4Validator] = {}
 
@@ -191,7 +191,14 @@ def validate_payload(message, ocpp_version):
     try:
         validator.validate(message.payload)
     except SchemaValidationError as e:
-        raise ValidationError(f"Payload '{message.payload} for action "
+        if (e.validator == SchemaValidators.type.__name__):
+            raise TypeConstraintViolationError(details={"cause":e.message})
+        elif (e.validator == SchemaValidators.additionalProperties.__name__):
+            raise FormatViolationError(details={"cause":e.message})
+        elif (e.validator == SchemaValidators.required.__name__):
+            raise ProtocolError(details={"cause":e.message})
+        else:
+            raise ValidationError(f"Payload '{message.payload} for action "
                               f"'{message.action}' is not valid: {e}")
 
 
--- a/ocpp/messages.py
+++ b/ocpp/messages.py
@@ -192,14 +192,14 @@ def validate_payload(message, ocpp_version):
         validator.validate(message.payload)
     except SchemaValidationError as e:
         if (e.validator == SchemaValidators.type.__name__):
-            raise TypeConstraintViolationError(details={"cause":e.message})
+            raise TypeConstraintViolationError(details={"cause": e.message})
         elif (e.validator == SchemaValidators.additionalProperties.__name__):
-            raise FormatViolationError(details={"cause":e.message})
+            raise FormatViolationError(details={"cause": e.message})
         elif (e.validator == SchemaValidators.required.__name__):
-            raise ProtocolError(details={"cause":e.message})
+            raise ProtocolError(details={"cause": e.message})
         else:
             raise ValidationError(f"Payload '{message.payload} for action "
-                              f"'{message.action}' is not valid: {e}")
+                                  f"'{message.action}' is not valid: {e}")
 
 
 class Call:
--- a/tests/test_messages.py
+++ b/tests/test_messages.py
@@ -7,6 +7,7 @@ from ocpp.v16.enums import Action
 from ocpp.exceptions import (ValidationError, ProtocolError,
                              FormatViolationError,
                              PropertyConstraintViolationError,
+                             TypeConstraintViolationError,
                              UnknownCallErrorCodeError)
 from ocpp.messages import (validate_payload, get_validator, _validators,
                            unpack, Call, CallError, CallResult, MessageType,
@@ -163,10 +164,10 @@ def test_validate_payload_with_valid_payload(ocpp_version):
     validate_payload(message, ocpp_version=ocpp_version)
 
 
-def test_validate_payload_with_invalid_payload():
+def test_validate_payload_with_invalid_additional_properties_payload():
     """
-    Test if validate_payload raises ValidationError when validation of
-    payload failes.
+    Test if validate_payload raises FormatViolationError when validation of
+    payload with unrequested properties fails.
     """
     message = CallResult(
         unique_id="1234",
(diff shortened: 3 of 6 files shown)


OrangeTux:
Hello @joaomariord, thanks for the PR! Although it's a backwards incompatible change,  I think it's a good improvement. 
Can you make sure to fix the lint errors?

joaomariord:
Hello @OrangeTux, thank you for promptly answering the PR, I've corrected the lint errors and adapted tests for the new behavior.

I think we could keep the NotImplemented exception for core ("required") OCPP routes, and only answer with CALLERROR-NotImplemented for the optional routes. I can do a PR on this if you agree.

Another thing that would be nice is to allow the programmer to give is own exception and error handling routines. I think the decorator could be used set the error handling callback (e.g. @on(Action.Heartbeat, my_error_handler)). An alternative, more complete, way would be to allow the routing of the errors to a given error handle (e.g. @error(Action.Hearbeat, TypeConstraintError)). 
These are just suggestions where I would like some insight on, I can put time to this if its worthy and makes sense.

joaomariord:
Hello @OrangeTux, thank you for promptly answering the PR, I've corrected the lint errors and adapted tests for the new behavior.

I think we could keep the NotImplemented exception for core ("required") OCPP routes, and only answer with CALLERROR-NotImplemented for the optional routes. I can do a PR on this if you agree.

Another thing that would be nice is to allow the programmer to give is own exception and error handling routines. I think the decorator could be used set the error handling callback (e.g. @on(Action.Heartbeat, my_error_handler)). An alternative, more complete, way would be to allow the routing of the errors to a given error handle (e.g. @error(TypeConstraintError, )). 
These are just suggestions where I would like some insight on, I can put time to this if its worthy and makes sense.

OrangeTux:
> Another thing that would be nice is to allow the programmer to give is own exception and error handling routines. I think the decorator could be used set the error handling callback (e.g. @on(Action.Heartbeat, my_error_handler)). An alternative, more complete, way would be to allow the routing of the errors to a given error handle (e.g. @error(Action.Hearbeat, TypeConstraintError)).

Thanks for the feedback. The current routing system works fine for most cases, but is limited if your the use case is more complex. You give a good example of one of the routing system's limitation: one is unable to control the response of a call if any exception is raised in this library. #295 highlights another limitation.

With your suggestion you gave me some food for though.

jainmohit2001:
>An alternative, more complete, way would be to allow the routing of the errors to a given error handle (e.g. @error(Action.Hearbeat, TypeConstraintError)). These are just suggestions where I would like some insight on, I can put time to this if its worthy and makes sense.

This would actually help a lot. I am trying to build a fully-fledged CSMS and this is much needed to handle all possible real-case scenarios.

jainmohit2001:
After reading both the documentation for OCPP-1.6 and OCPP-2.0.1, I found that the errors are based on whether the payload received by the CSMS follows the RPC framework or whether the CP sends a CALL which is not a BootNotification request or a message triggered by TriggerMessage request 

@joaomariord handled the first case elegantly. 
And for the second one, I have to apply a check inside all the `@on` decorated functions for the CSMS as well as on the CP side as mentioned in #295 .

If it's possible to create a 'catch all' error handler that knows from which `@on` decorated function this error is generated and what is the error type, e.g. SecurityError, then we can leave the implementation up to the user. This can help the user customize the actions it has to take on some specific set of errors generated and by which function it is being raised.

joaomariord:
@jainmohit2001 I completely agree with your idea of a 'catch all' error handler.
If we can have an umbrella error handler, it seems there is no need for handling each specific error with a unique handler, as we can also do this in the umbrella handler.
There is only one quirk about that, maybe it is more pythonic to have something like:
```
@on(Action.Hearbeat)
def heartbeat_handler():
    ....

@error(Action.Heartbeat, ProtocolError)
def protocol_error_handler(error: ProtocolError):
   .....
   return True # On Error handled
   return False # On not handled
```
instead of:

```
def ProtocolErrorHandler(error: ProtocolError): -> bool 
   
   return True # On Error handled
   return False # On not handled

def heartbeat_error_handler(error: BaseException | Any):
   sub_handler = {
           ProtocolError.__class__: ProtocolErrorHandler,
           ... #Other Handlers
   }.get(error.__class__, None)   
   if sub_handler in not None:
       return sub_handler()
   else:
       return False

@on(Action.Hearbeat, heartbeat_error_handler)
def heartbeat_handler():
    ....
```

The "switch-case" is just a quick and dirty way to do it, but ignore it for the sake of the example, this could be solved in many more elegant ways.

jainmohit2001:
I actually thought of something else though. 
Can we have a default error handler in the ChargePoint class that does nothing and is called whenever a `@on` method raises an error? This method can be overridden by the user itself if the user wishes to handle the errors.

joaomariord:
That should be easy enough, it's an elegant solution.


-------- Original message --------
From: Mohit Jain ***@***.***>
Date: 1/28/22 11:59 (GMT+00:00)
To: mobilityhouse/ocpp ***@***.***>
Cc: João Mário Domingos ***@***.***>, Mention ***@***.***>
Subject: Re: [mobilityhouse/ocpp] Improved error handling during schema validation (PR #294)


I actually thought of something else though.
Can we have a default error handler in the ChargePoint class that does nothing and is called whenever a @on method raises an error? This method can be overridden by the user itself if the user wishes to handle the errors.

—
Reply to this email directly, view it on GitHub<https://emea01.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fmobilityhouse%2Focpp%2Fpull%2F294%23issuecomment-1024147383&data=04%7C01%7C%7C1c4b89c7807f46cd573d08d9e255a5b2%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C637789679751114554%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000&sdata=QnuPlBkiOHXGX00Xyih6yevp0ByG35gQoS0%2FjoR9U2s%3D&reserved=0>, or unsubscribe<https://emea01.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fnotifications%2Funsubscribe-auth%2FAGGSOG67F373IRY667732C3UYKAKJANCNFSM5MZKZ2VA&data=04%7C01%7C%7C1c4b89c7807f46cd573d08d9e255a5b2%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C637789679751114554%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000&sdata=F3x49DegVMjVN5o0Uqb9Hm8MLnPLEA11UUn0V9Z3RA0%3D&reserved=0>.
Triage notifications on the go with GitHub Mobile for iOS<https://emea01.safelinks.protection.outlook.com/?url=https%3A%2F%2Fapps.apple.com%2Fapp%2Fapple-store%2Fid1477376905%3Fct%3Dnotification-email%26mt%3D8%26pt%3D524675&data=04%7C01%7C%7C1c4b89c7807f46cd573d08d9e255a5b2%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C637789679751114554%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000&sdata=fA9ujgDSQrlbATszGCgnAPbHHtgt27CrzmuUbhoEjk8%3D&reserved=0> or Android<https://emea01.safelinks.protection.outlook.com/?url=https%3A%2F%2Fplay.google.com%2Fstore%2Fapps%2Fdetails%3Fid%3Dcom.github.android%26referrer%3Dutm_campaign%253Dnotification-email%2526utm_medium%253Demail%2526utm_source%253Dgithub&data=04%7C01%7C%7C1c4b89c7807f46cd573d08d9e255a5b2%7C84df9e7fe9f640afb435aaaaaaaaaaaa%7C1%7C0%7C637789679751114554%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000&sdata=17jhsIcsfZV2TXK4q61p7wYx8steo%2BA6Jl7t83F2Kpg%3D&reserved=0>.

jainmohit2001:
Yeah something like this, but where can we integrate this in the code?

I think the following should preserve the backward compatibility

```python
# ocpp.charge_point.py
try:
    if inspect.isawaitable(response):
        response = await response
except Exception as e:
    LOGGER.exception("Error while handling request '%s'", msg)
    response = msg.create_call_error(e).to_json()
    await self._send(response)

    # await self.on_exception() # Pass the relevant args and kwargs

    return
```

jainmohit2001:
>We might want to change the name of `on_exception()` to something more specific. `on_exceptions_while_handling_calls()` >would be very specific, but also a bit verbose. Not sure if you've better suggestion.

Yeah, I totally agree with you here. THe name of the function should reflect that it is called after the CALLERROR is sent and the user wants to do something more.

proelke:
Personally I like the proposed idea of an error decorator rather than a single function. A single function will end up with a rather long set of isinstance() statements depending on how many messages have error handling implemented for.

proelke:
I think it could be added quite quickly and I can't imagine off hand how it could break backwards compatibility. It could be designed based on the after decorator.

jainmohit2001:
> I think it could be added quite quickly and I can't imagine off hand how it could break backwards compatibility. It could be designed based on the after decorator.

If it's possible, please include this funtionality. @joaomariord @OrangeTux @proelke .

joaomariord:
@OrangeTux can you review this PR? It would be great to have this merged by the end of the week.

I think that what we discussed needs to be further developed and tested, and in that sense this PR could advance right away.

OrangeTux:
I've a last suggestion that combines several of the discussed variants in one.

What if we introduce a decorater `@after_exception()` with the following signature:

``` python
class Charger(CP):
    @after_exception()
    async def on_exception(self, call: Call, exc: Exception) -> None:
         """ Override that allows for custom error handling. """
         if isinstance(call, call.Heartbeat):
             // do something
```
The handler has similar semantics as the `@after()` decorator. It receives the `Call` that caused the exception and the exception itself. Similar to `@after()` any return value is ignored and any exception raised in the handler is ignored too. 
Similar to `@after()`, `@after_exception()` is executed after the CallError has been send to the other peer. In other words, this handler can't influence that response.

I choose the decorator approach over the approach of overriding a method. The former is consistent with `@after()` and allows for adding multiple handlers in the future.

In a future release we could implement filtering by passing arguments to `@after_exception()` so that we can call it like `@after_exception(action = 'Heartbeat' | 'MeterValue')` or `@after_exception(exc_type=ProtocolError)`.
But before we do that, I'm curious to see how this API holds up. Even without these arguments to filter for specific calls or exceptions, one can implement such filtering themself. 

The change is backwards compatible.

If an exception is raised in the `@on()` handler, this `@after_exception()` is called. The regular `@after()` should not be called.

@joaomariord  I have no plans to work on this topic this week, but I'm happy to receive a PR.

joaomariord:
> @joaomariord I have no plans to work on this topic this week, but I'm happy to receive a PR.
I was talking about the current PR. If I have the time I can try to work our conversation into an actual PR.

OrangeTux (approved):

joaomariord:
Hi @OrangeTux, @jainmohit2001, and @proelke, I've not been available for the past few weeks, but have been passing my eyes through the code and something came up.
Could this be done with the current state of the code?

Would something like:
- SkipSchemaValidation on @on,
- A call to validate_payload(snake_to_camel_case(payload)) inside the on_action handler method
- Wrap that call inside try except

, work?

PR #299, opened 2022-02-03
Add TxStartStopPoint enum

mdwcrft:
Adds TxStartStopPoint as an enum. Although this isn't an "EnumType" in the spec it similarly is a list of allowed values.

If this should go somewhere other than `enums.py` let me know but thought it would be useful to have it somewhere in this lib.

Screenshot 2022-02-03 at 15 29 31

(page 434 of spec part 2)


tropxy (changes requested):
LGTM, could you add tests for this? We dont have any for the enums of v201, but you can see what was done in v16: https://github.com/mobilityhouse/ocpp/blob/master/tests/v16/test_enums.py

OrangeTux (approved):
Thanks!

Issue #302, opened 2022-02-04
v201/Wrong dataclass MeterValueType


Shadowsith:
MeterValueType was implemented correctly in the JSON schema but the dataclass is implemented incorrectly.

current (wrong):
``` python
@dataclass
class MeterValueType:
    """
    Collection of one or more sampled values in MeterValuesRequest and
    TransactionEvent. All sampled values in a MeterValue are sampled at the
    same point in time.
    MeterValueType is used by: MeterValuesRequest, TransactionEventRequest
    """

    timestamp: str
    sampled_value: SampledValueType
```

**should:**
``` python
@dataclass
class MeterValueType:
    """
    Collection of one or more sampled values in MeterValuesRequest and
    TransactionEvent. All sampled values in a MeterValue are sampled at the
    same point in time.
    MeterValueType is used by: MeterValuesRequest, TransactionEventRequest
    """

    timestamp: str
    sampled_value: List[SampledValueType]
```

See **MeterValueType** specification https://raw.githubusercontent.com/mobilityhouse/ocpp/master/docs/v201/OCPP-2.0.1_part2_specification.pdf page 381, **sampledValue** (1...*)

OrangeTux:
Thanks for reporting the issue.

Shadowsith:
I've created a PR for that: https://github.com/mobilityhouse/ocpp/pull/306

Jared-Newell-Mobility:
This was merged into master some time ago, so I will close this

PR #306, opened 2022-02-07
Fixed MeterValueType sampled_value attribute type

Shadowsith:
Fixed issue: https://github.com/mobilityhouse/ocpp/issues/302


OrangeTux (approved):
Thanks for the PR!

OrangeTux (changes requested):
`List` is not brought into the scope. Can you fix that?

PR #316, opened 2022-03-04
Update datatypes.py

HugoJP1:
Hi - I was attempting to use `GetVariableResultType`, and realised that there is a typo in the definition of `variable`.

### DIFF
--- a/ocpp/v201/datatypes.py
+++ b/ocpp/v201/datatypes.py
@@ -436,7 +436,7 @@ class GetVariableResultType:
 
     attribute_status: enums.GetVariableStatusType
     component: ComponentType
-    variable = VariableType
+    variable: VariableType
     attribute_type: Optional[enums.AttributeType] = None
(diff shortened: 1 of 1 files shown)


OrangeTux (approved):
Good catch! Thanks for fixing.

OSkrk:
Hello @OrangeTux - How do you generate the datatypes.py file from the schemas?

OrangeTux:
We created a few scripts that generate this code based on the JSON schema. You can find them in https://github.com/mobilityhouse/ocpp/tree/master/scripts . Note that the output is not without errors and require manual fix up.

OSkrk:
Thank you for your answer @OrangeTux, I have tested the scripts and succesfully generated call, call_result and enums, but the file datatypes.py is not generated by these scripts. Is there a missing script?

OrangeTux:
I'm afraid we did that manually...

PR #335, opened 2022-05-10
Improve Exception handling and CallError responses

proelke:
This PR is to improve the exception handling in validating messages and to make the CallError messages closer to the OCPP spec. Quick overview:

- Added missing `NotSupported` to the exception list.
- Changed usage of `NotImplemented` to `NotSupported` when a handler doesn't exist.
- Use the default description and moved additional information to `details`
- Reduce the usage of `ValidationError` as it is not a valid OCPP error type, doesn't generate a CallError, and crashes the code when handling messages. I've tried to replace them with relevant OCPP error types.

### DIFF
--- a/ocpp/charge_point.py
+++ b/ocpp/charge_point.py
@@ -9,7 +9,7 @@ from dataclasses import asdict
 
 from ocpp.routing import create_route_map
 from ocpp.messages import Call, validate_payload, MessageType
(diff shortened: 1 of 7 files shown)


OrangeTux (approved):
Thanks. Adding details to the CallError is pretty useful for debugging.

proelke:
❤️

PR #336, opened 2022-05-11
Handle responder_url edge case

proelke:
In the 2.0.1 spec for `AuthorizeRequest` and `GetCertificateStatusRequest` there is a field responderURL that does not get converted from snake_case `responder_url` to camelCase `responderURL`. This PR adds handling of this edge case. in `SetNetworkProfileRequest` there is a field `ocppCsmsUrl` which is handled correctly but causes problems for a more general solution.

Let me know if the conditionals work for you.


tropxy (approved):
needs rebasing

OrangeTux:
@Jared-Newell-Mobility This one is also required for a v1 release. You can take it over the issue and finish it.

Jared-Newell-Mobility:
Thanks for highlighting this, it has been replaced with PR https://github.com/mobilityhouse/ocpp/pull/592, so it will close for now

PR #369, opened 2022-09-13
Make CI use latest bugfix release of Python 3.10

adamchainz:
Pinning to `3.10.0` was probably done to avoid using `3.10`, which is a YAML float with vlaue `3.1`. Instead of doing that, use a string specifier, which installs the latest bugfix release of Python 3.10 (currently 3.10.7). This will stop any possible encounters with already-fixed bugs in Python.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -10,7 +10,11 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        version: [3.7, 3.8, 3.9, 3.10.0]
+        version:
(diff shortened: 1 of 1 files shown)


OrangeTux (approved):
Thanks!

adamchainz:
GitHub branch protection needs updating for the change in checks, as this PR is currently showing:

>  tests (3.10.0) Expected — Waiting for status to be reported
>  tests (3.6) Expected — Waiting for status to be reported 

I guess 3.6 has been removed, and it's this PR that renames the 3.10.0 check to 3.10.

PR #384, opened 2022-11-14
[DVI-509] OCPP Validation logging

klimaschkas:
Scope: In case an OCPP validation fails TypeConstraint or Format, we should log which message caused the validation to fail.

We noticed a reproducible TypeConstraintViolation as a reaction to an OCPP message from Etrel Chargers.

A (OCPP) response from the Etrel chargers leads to a a) problem in data type conversion/parsing, b) a debug message that a transmitted string may be too long.

### DIFF
--- a/ocpp/messages.py
+++ b/ocpp/messages.py
@@ -68,13 +68,15 @@ def unpack(msg):
         msg = json.loads(msg)
     except json.JSONDecodeError:
         raise FormatViolationError(
-            details={"cause": "Message is not valid JSON"})
+            details={"cause": "Message is not valid JSON",
+                     "ocpp_message": msg})
 
     if not isinstance(msg, list):
         raise ProtocolError(
             details={"cause": ("OCPP message hasn't the correct format. It "
                                f"should be a list, but got '{type(msg)}' "
-                               "instead")})
+                               "instead"),
+                     "ocpp_message": msg})
 
     for cls in [Call, CallResult, CallError]:
         try:
@@ -82,13 +84,15 @@ def unpack(msg):
                 return cls(*msg[1:])
         except IndexError:
             raise ProtocolError(
-                details={"cause": "Message does not contain MessageTypeId"})
+                details={"cause": "Message does not contain MessageTypeId",
+                         "ocpp_message": msg})
         except TypeError:
             raise ProtocolError(
-                details={"cause": "Message is missing elements."})
+                details={"cause": "Message is missing elements.",
+                         "ocpp_message": msg})
 
     raise PropertyConstraintViolationError(
-        details={f"MessageTypeId '{msg[0]}' isn't valid"})
+        details={"cause": f"MessageTypeId '{msg[0]}' isn't valid"})
 
 
 def pack(msg):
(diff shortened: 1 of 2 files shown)


ocpp/messages.py
```diff
@@ -360,43 +319,134 @@ class CallError:
...
+                message.message_type_id, message.action, ocpp_version
+            )
+    except (OSError, json.JSONDecodeError):
+        raise NotImplementedError(
+            details={"cause": f"Failed to validate action: {message.action}"}
+        )
+
+    try:
+        validator.validate(message.payload)
+    except SchemaValidationError as error:
```
klimaschkas, line 430: Most of the changes come from running black, the main changes were extending the error details with the ocpp_message in this part of the code

tmh-grunwald-markus:
Please add a meaningful description for the PR. Copy/Pasting from the Jira issue is allowed. Someone who was not involved should easily understand what the PR addresses and how that is implemented.

Also, for the next time: If most of the changes come from applying flake8/black it would be nice to have those changes in a separate commit and the "real" payload in another. `git add --patch` is the way to do that.

tmh-grunwald-markus:
Would it make sense to add the `message` to this exception, too?

https://github.com/mobilityhouse/ocpp/blob/DVI-509-validation-logging/ocpp/messages.py#L376-L380

tests/test_exceptions.py
```diff
@@ -1,15 +1,62 @@
...
+    )
+
+    regex_pattern = (
+        r"'ocpp_message': <Call - unique_id=123456, action=BootNotification, "
+        r"payload={'chargePointVendor': 1, "
+        r"'chargePointModel': 'SingleSocketCharger'}"
+    )
+
+    with pytest.raises(TypeConstraintViolationError, match=regex_pattern):
+        validate_payload(call, "1.6")
```
tmh-grunwald-markus, line 43: Nitpick:

Is is, technically, a regex pattern. But as far as I can see it is a trivial one. I would rename it to `pattern`, because then reviewers don't start to look for (real) regex patterns.

Since parsing regular expressions is time consuming, you could just use `pytest.raises` like this:
https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions

Both suggestions are completely optional and more a matter of taste.

ocpp/messages.py
```diff
@@ -360,43 +319,134 @@ class CallError:
...
-               f"error_details={self.error_details}>"
+        return (
+            f"<CallError - unique_id={self.unique_id}, "
+            f"error_code={self.error_code}, "
+            f"error_description={self.error_description}, "
+            f"error_details={self.error_details}>"
+        )
+
+
+def validate_payload(message: Union[Call, CallResult],
```
tmh-grunwald-markus, line 372: Only out of curiosity: Why did you move the function here from its original position?

tmh-grunwald-markus (changes requested):
The only requested change: Please add a meaningful description for the PR. Copy/Pasting from the Jira issue is allowed. Someone who was not involved should easily understand what the PR addresses and how that is implemented.
A good description also helps to get a good commit message for the squash merge.

The remaining comments are just ... comments and a little nitpicky :)

OrangeTux:
A PR just got merged that introduces black and isort. Can you run these tools and update you're branch?

ocpp/messages.py
klimaschkas: To use the type hint for the message, which refers to the Call and CallResult class, the function has to be after the class definitions

tmh-grunwald-markus: I see. You could use this import to postpone evaluation of the type hint/annotations, without moving code around:

```
from __future__ import annotations
```

tmh-grunwald-markus (approved):
Nice :)

OrangeTux (approved):

PR #388, opened 2022-11-16
feat: Support OCPP spec extensions - Part 1

proelke:
This PR adds support for OCPP spec extensions. This is the first PR adding support for V2X extensions but also enabling support for potential future extensions from the OCA. The V2X extension types, schemas, examples, and unit tests would come in a future PR.

### DIFF
--- a/ocpp/charge_point.py
+++ b/ocpp/charge_point.py
@@ -178,7 +178,11 @@ class ChargePoint:
             )
 
         if not handlers.get("_skip_schema_validation", False):
-            validate_payload(msg, self._ocpp_version)
+            extension = handlers.get("_extension")
+            if extension:
+                validate_payload(msg, self._ocpp_version, extension=extension)
+            else:
+                validate_payload(msg, self._ocpp_version)
         # OCPP uses camelCase for the keys in the payload. It's more pythonic
         # to use snake_case for keyword arguments. Therefore the keys must be
         # 'translated'. Some examples:
--- a/ocpp/charge_point.py
+++ b/ocpp/charge_point.py
@@ -5,10 +5,11 @@ import re
 import time
 import uuid
 from dataclasses import asdict
+from importlib import import_module
 from typing import Dict, List, Optional, Union
 
 from ocpp.exceptions import NotSupportedError, OCPPError
-from ocpp.messages import Call, Extension, MessageType, unpack, validate_payload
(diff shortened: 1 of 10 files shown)


ocpp/charge_point.py
```diff
@@ -290,15 +303,18 @@ async def call(self, payload, suppress=True):
...
         # Create the correct Payload instance based on the received payload. If
         # this method is called with a call.BootNotificationPayload, then it
         # will create a call_result.BootNotificationPayload. If this method is
         # called with a call.HeartbeatPayload, then it will create a
         # call_result.HeartbeatPayload etc.
-        cls = getattr(self._call_result, payload.__class__.__name__)  # noqa
+        if extension:
+            cls = getattr(extension_call_result, payload.__class__.__name__)  # noqa
+        else:
+            cls = getattr(self._call_result, payload.__class__.__name__)  # noqa
```
proelke, line 317: @OrangeTux I feel there is probably a better way to handle this situation, but this is where I landed. I am open to suggestions to improve this.

proelke:
As I cannot stack the PRs under the TMH repo from the fork, here are the follow up PRs:

https://github.com/proelke/ocpp/pull/1
https://github.com/proelke/ocpp/pull/2

OrangeTux (commented):
Thanks for your PRs.

I think we should try to modify the `ChargePoint.call()` and `@route()` API in such way, that they allow for custom validations. Support for any OCA extension can be build on top of that API.

What about introducing a new object, let's call it `Validator`. 

A `Validator` implements two methods:

``` python
class Validator:
   def __init__(self, path_to_schemas: str):
       ....

   def get_schema_for_call(self, action: str):
        ...
    
   def get_schema_for_call_response(self, action: str):
        ...
```

We could implement `Validator`s for OCPP extensions. But users can create their own `Validator` when they've custom extensions.

Assume a `Validator` for V2X:

``` python
// ocpp.v201.extensions.v2x.py
validator = Validator('path/to/folder/with/schemas')
```

Now one can use the validator like such:


``` python
from ocpp.v201.extensions.v2x

@route('NotifyEVChargingNeeds', validator=v2x.validator)
async def on_notify_ev_charging_needs(*args, **kwargs):
   ...
```
I'm not sure if I like the names I picked, but I hope you get the idea.

proelke:
@OrangeTux I updated the PR based on your feedback.

PR #407, opened 2023-01-04
Fix publish-to-pypi build

OrangeTux:
[The publish-to-pypi build failed for 0.16.0.](https://github.com/mobilityhouse/ocpp/actions/runs/3838953465/jobs/6536318630). It seems that the issue is caused by the lack of quotes around value for field 'python-version'.


tmh-azinhal (approved):
Neat :ok_hand:

PR #413, opened 2023-01-24
Configuration key enums

isabelle-tmh:
Follow up of https://github.com/mobilityhouse/ocpp/pull/231 PR with the addition of new `ConfigurationKey`s from OCPP 1.6 ISO 15118 v10 and security whitepaper edition.


tonirosemann (commented):
Looks good, but i guess we need @OrangeTux for this
didn't check the correctnes of each entry though

OrangeTux (approved):

PR #426, opened 2023-03-06
minor fixes

drc38:
According to the notes I assume this is incorrect and should be NotImplemented?


PratikBodawala (approved):

tropxy (approved):

drc38:
@OrangeTux a couple of small fixes for review

OrangeTux:
Table 7 of the Open Charge Point Protocol JSON 1.6, OCPP-J 1.6 Specification lists all error codes with a short description. 

* NotSupported -  Requested Action is not known by receiver
* NotImplemented - Requested Action is recognized but not supported by the receiver `charge_point.py:175` can indicate both problems. 

Trying to send a request with action "InvalidAction" triggers this code path. In this case, a `NotSupportedError` would be fitting. 

However, the code path is also executed when the receiver receives a request for an action the receiver doesn't implement. For example, if a charger receives a "GetDiagnostics" request without supporting this feature.  Now `NotImplementedError` would be suitable.

A while ago, we changed the error type from "NotImplementedError` to `NotSupportedError`. See https://github.com/mobilityhouse/ocpp/commit/c6e7e484f4cc263fb15c5f0802e70fe8eced1944

I'm hesitant to revert that change by approving this PR. Instead, some code should be include to distinguish between 2 scenarios I described and raise the appropriate error.

drc38:
> Table 7 of the Open Charge Point Protocol JSON 1.6, OCPP-J 1.6 Specification lists all error codes with a short description.
> 
> * NotSupported -  Requested Action is not known by receiver
> * NotImplemented - Requested Action is recognized but not supported by the receiver `charge_point.py:175` can indicate both problems.
> 
> Trying to send a request with action "InvalidAction" triggers this code path. In this case, a `NotSupportedError` would be fitting.
> 
> However, the code path is also executed when the receiver receives a request for an action the receiver doesn't implement. For example, if a charger receives a "GetDiagnostics" request without supporting this feature. Now `NotImplementedError` would be suitable.
> 
> A while ago, we changed the error type from "NotImplementedError`to`NotSupportedError`. See [c6e7e48](https://github.com/mobilityhouse/ocpp/commit/c6e7e484f4cc263fb15c5f0802e70fe8eced1944)
> 
> I'm hesitant to revert that change by approving this PR. Instead, some code should be include to distinguish between 2 scenarios I described and raise the appropriate error.

Cheers, I wasn't aware of the history. I have reverted and amended the comment to align with the choice of using the NotSupportedError.

Jared-Newell-Mobility:
@drc38 - I see that the change in the comment field is valid, could you please recreate this PR, having it branched off master, thanks?

PR #427, opened 2023-03-06
Add py.typed file

roguh:
This enables type checkers like [mypy](https://www.mypy-lang.org/) to look into the library's source code to determine the types of many values. See https://dev.to/whtsky/don-t-forget-py-typed-for-your-typed-python-package-2aa3

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,215 +1,223 @@
 # Change log
 
+## 0.19.0 (2023-04-26)
+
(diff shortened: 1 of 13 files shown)


roguh:
I'm adding additional type hints here: https://github.com/edf-re/types-ocpp

roguh:
Seems this is in main already. Thanks!

rick-jennings:
Hi, I don't see the py.typed file and am having issues with type hints on version 0.19.0 in VS Code.  

Is there a plan to add py.typed file in the future to support type hints?  Or may I be missing something?  Thanks!

PR #438, opened 2023-04-13
feat: add optional param for passing an unique_id to call method.

santiagosalamandri:
This PR adds:
-An additional optional parameter to the `call` method. 

This is needed for requirement E-13 on OCPP2.0.1 when the CS needs to retry sending a transaction message with the same ID.
Currently, when the `call` method is called, a `Call` object is created with a new random ID.

### DIFF
--- a/ocpp/charge_point.py
+++ b/ocpp/charge_point.py
@@ -237,7 +237,7 @@ class ChargePoint:
             # when no '_on_after' hook is installed.
             pass
 
-    async def call(self, payload, suppress=True):
+    async def call(self, payload, suppress=True, unique_id=None):
         """
         Send Call message to client and return payload of response.
 
@@ -261,8 +261,12 @@ class ChargePoint:
         """
         camel_case_payload = snake_to_camel_case(asdict(payload))
 
+        unique_id = (
+            unique_id if unique_id is not None else str(self._unique_id_generator())
+        )
+
         call = Call(
-            unique_id=str(self._unique_id_generator()),
+            unique_id=unique_id,
             action=payload.__class__.__name__[:-7],
             payload=remove_nones(camel_case_payload),
         )
--- a/tests/v16/conftest.py
+++ b/tests/v16/conftest.py
@@ -40,8 +40,8 @@ def base_central_system(connection):
 @pytest.fixture
 def mock_boot_request():
     return call.BootNotificationPayload(
-        charge_point_vendor="ICU Eve Mini",
-        charge_point_model="ICU Eve Mini",
+        charge_point_vendor="dummy_vendor",
+        charge_point_model="dummy_model",
     )
 
 
--- a/tests/v16/conftest.py
+++ b/tests/v16/conftest.py
@@ -1,4 +1,9 @@
-from unittest.mock import AsyncMock
+try:
+    from unittest.mock import AsyncMock
+except ImportError:
+    # Python 3.7 and below don't include unittest.mock.AsyncMock. Hence,
+    # we need to resolve to a package on pypi.
+    from asynctest import CoroutineMock as AsyncMock
 
 import pytest
 
--- a/tests/v16/conftest.py
+++ b/tests/v16/conftest.py
@@ -1,7 +1,9 @@
+from unittest.mock import AsyncMock
+
 import pytest
 
-from ocpp.messages import Call
-from ocpp.v16 import ChargePoint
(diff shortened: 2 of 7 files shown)


OrangeTux (changes requested):
Thanks for the PR. 

Can you add or modify a unit test to cover the change?

santiagosalamandri:
@OrangeTux  I added some tests that cover this new functionality. To maintain consistency with the existing codebase, I followed the same approach used in the repository and duplicated the same tests for the different OCPP versions.
However, I'm currently receiving an error during code analysis regarding duplicated code, do you have any advice for this?

santiagosalamandri:
> @santiagosalamandri not sure this is a SonarCloud bug (it happened in the past) but can you try to reduce duplication by using parametrized python testing?
> 
> For example, this code:
> 
> ```python
> @pytest.mark.asyncio
> async def test_call_with_unique_id_should_return_same_id(
>     mock_boot_request, mock_base_central_system
> ):
> 
>     expected_unique_id = "12345"
>     # Call the method being tested with a unique_id as a parameter
>     await mock_base_central_system.call(mock_boot_request, unique_id=expected_unique_id)
>     (
>         actual_unique_id,
>         _,
>     ) = mock_base_central_system._get_specific_response.call_args_list[0][0]
> 
>     # Check the actual unique id is equals to the one passed to the call method
>     assert actual_unique_id == expected_unique_id
> 
> 
> @pytest.mark.asyncio
> async def test_call_without_unique_id_should_return_a_random_value(
>     mock_boot_request, mock_base_central_system
> ):
> 
>     expected_unique_id = str(mock_base_central_system._unique_id_generator())
> 
>     # Call the method being tested without passing a unique_id as a parameter
>     await mock_base_central_system.call(mock_boot_request)
> 
>     (
>         actual_unique_id,
>         _,
>     ) = mock_base_central_system._get_specific_response.call_args_list[0][0]
>     # Check the actual unique id is equals to the one internally generated
>     assert actual_unique_id == expected_unique_id
> ```
> 
> could be reduced to one test with the `unique_id` being parametrized as argument for the test.

Hi @tropxy , yes, that will reduce the code duplication but it won't reduce the percentage duplication(which has to be less than 3% to pass), as it is calculated based on new code among the files that have changed. So, to overcome this, I could create different PRs for each test, what do you think @OrangeTux ?

tropxy (approved):

tropxy:
> > @santiagosalamandri not sure this is a SonarCloud bug (it happened in the past) but can you try to reduce duplication by using parametrized python testing?
> > For example, this code:
> > ```python
> > @pytest.mark.asyncio
> > async def test_call_with_unique_id_should_return_same_id(
> >     mock_boot_request, mock_base_central_system
> > ):
> > 
> >     expected_unique_id = "12345"
> >     # Call the method being tested with a unique_id as a parameter
> >     await mock_base_central_system.call(mock_boot_request, unique_id=expected_unique_id)
> >     (
> >         actual_unique_id,
> >         _,
> >     ) = mock_base_central_system._get_specific_response.call_args_list[0][0]
> > 
> >     # Check the actual unique id is equals to the one passed to the call method
> >     assert actual_unique_id == expected_unique_id
> > 
> > 
> > @pytest.mark.asyncio
> > async def test_call_without_unique_id_should_return_a_random_value(
> >     mock_boot_request, mock_base_central_system
> > ):
> > 
> >     expected_unique_id = str(mock_base_central_system._unique_id_generator())
> > 
> >     # Call the method being tested without passing a unique_id as a parameter
> >     await mock_base_central_system.call(mock_boot_request)
> > 
> >     (
> >         actual_unique_id,
> >         _,
> >     ) = mock_base_central_system._get_specific_response.call_args_list[0][0]
> >     # Check the actual unique id is equals to the one internally generated
> >     assert actual_unique_id == expected_unique_id
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > could be reduced to one test with the `unique_id` being parametrized as argument for the test.
> 
> Hi @tropxy , yes, that will reduce the code duplication but it won't reduce the percentage duplication(which has to be less than 3% to pass), as it is calculated based on new code among the files that have changed. So, to overcome this, I could create different PRs for each test, what do you think @OrangeTux ?

Yup, understood, on the light of this I think we can just ignore the sound cloud analysis. All tests pass, which is good.

PR #451, opened 2023-07-17
Add code generator for OCPP 2.1

OrangeTux:
This PR introduces support for OCPP 2.1. The code in ocpp/v21 is generated by the newly
introduced code generator. 

I've created this draft to gather feedback on the generated code in ocpp/v21, not on the code generator itself.
The generator requires some clean up and tests.

Unfortunately, I'm not allowed to include the JSON schemas with this PR at the moment. The OCA
distributed these files to members only. I'm in touch with the OCA to see if they allow me to include
the schemas in this project.

Fixes: #452


Jared-Newell-Mobility:
I think we should remove from the PR the calls and calls results .json for now - maybe only include the generator components and then I can read in the schemas and compare the output results, then give feed back if it comes out correctly?

alexmclarty:
It is frustrating that the OCA doesn't allow these artefacts to be in the public domain! I know this isn't the place to discuss this, so I'll not go on for too long... 😁 

It's the same for the OCPP docs. They're available in PDF, not HTML, which makes creating tools that link to specific sections tricky. You have to add a reference to, for example, `5.16. Set Charging Profile`. 

It'd be 🆒 to get these specifications in the public domain and reduce the friction for contribution and innovation.

Issue #471, opened 2023-09-20
ImportError with jsonschema==v4.19.1


rasaffie:
With the new release of [jsonschema v4.19.1](https://github.com/python-jsonschema/jsonschema/compare/v4.19.0...v4.19.1) there is an `ImportError` on [ocpp/messages.py#12](https://github.com/mobilityhouse/ocpp/blob/master/ocpp/messages.py#L12).

The library renamed the variable `_validators` to `_keywords`, hence the `ImportError`.

Possible fixes are to rename the import or set the dependency to jsonschema==v4.19.0.

Also, is there a safer way to implement the functionality without importing a private variable of this library?

maiuka:
Hello there! We use this great lib for quite some time and now we have the same error. When you will fix it and publish to pypi?

Issue #447, opened 2023-06-30
Python 3.11 changes behavior how enums are formatted as string


OrangeTux:
@proelke made me aware about a changes to Python 3.11 that introduce [subtle changes to all enums](https://docs.python.org/3/whatsnew/3.11.html#enum) that subclass from both `str` and `enum.Enum`. Practically all enums defined in this library are subclassed from these 2 classes.

Consider the following code:

```python
import enum

class ResetType(str, enum.Enum):
    hard = "hard"
    soft = "soft"

# Passes in both Python 3.11 and earlier.
assert ResetType.hard == "hard"

# Passes in Python 3.10 and below, but fails in Python 3.11.
assert f'{ResetType.hard}' == "hard"
```

```
$  python3.11 /enum_test.py
Traceback (most recent call last):
  File "/tmp/enum_test.py", line 15, in <module>
    assert f'{ResetType.hard}' == "hard"
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

The underlying cause is that Python 3.11 changes the behavior of `enum.Enum.__format__()`  slightly. `f-strings` (and a few other string formatting methods) rely on dunder method `__format___()`.

So far, I considered two solutions. 

## Solution 1: StrEnum 
Python 3.11 introduces [`enum.StrEnum`](https://docs.python.org/3/library/enum.html#enum.StrEnum). One could implement a solution as described at https://tomwojcik.com/posts/2023-01-02/python-311-str-enum-breaking-change

```python
try:
    # breaking change introduced in python 3.11
    from enum import StrEnum
except ImportError:  # pragma: no cover
    from enum import Enum  # pragma: no cover

    class StrEnum(str, Enum):  # pragma: no cover
        pass  # pragma: no cover

class ResetType(StrEnum):
    hard = "hard"
    soft = "soft"

# Passes in both Python 3.11 and earlier.
assert ResetType.hard == "hard"

# Passes now in Python 3.10 and below and Python 3.11!
assert f'{ResetType.hard}' == "hard"
```

## Manually implement `__format__()`
 
The other solution is to implement `__format__()` manually. 
```python
import enum

class _Enum(str, enum.Enum):
    def __format__(self, spec) -> str:
        return str.__format__(str(self.name), spec)

class ResetType(_Enum):
    hard = "hard"
    soft = "soft"

# Passes in both Python 3.11 and earlier.
assert ResetType.hard == "hard"

[...]

PR #475, opened 2023-09-22
Add support for python3.11, drop support for python 3.7

rasaffie:
Fix issues https://github.com/mobilityhouse/ocpp/issues/471 and https://github.com/mobilityhouse/ocpp/issues/447


ocpp/v20/schemas/BootNotificationRequest_v1p0.json
```diff
@@ -31,7 +31,7 @@
           "maxLength": 20
         },
         "modem": {
-          "$ref": "#definitions/ModemType"
+          "$ref": "#/definitions/ModemType"
```
rasaffie, line 34: Change required do to `jsonschema-specifications` dependency.

tests/test_messages.py
```diff
@@ -261,17 +261,19 @@ def test_call_error_representation():
 def test_call_representation():
     call = Call(unique_id="1", action=Action.Heartbeat, payload={})
 
-    assert str(call) == "<Call - unique_id=1, action=Heartbeat, payload={}>"
+    assert str(call) == (
+        f"<Call - unique_id=1, action={Action.Heartbeat}" ", payload={}>"
```
rasaffie, line 265: * For `python >= 3.11`, `Action.Heartbeat` returns `"Action.Heartbeat"`.
* For `python < 3.11`, `Action.Heartbeat` returns `"Heartbeat"`.

tests/test_messages.py
```diff
@@ -261,17 +261,19 @@ def test_call_error_representation():
...
 def test_call_result_representation():
     call = CallResult(
         unique_id="1", action=Action.Authorize, payload={"status": "Accepted"}
     )
 
-    assert (
-        str(call) == "<CallResult - unique_id=1, action=Authorize, payload={'status': "
-        "'Accepted'}>"
+    assert str(call) == (
+        f"<CallResult - unique_id=1, action={Action.Authorize}"
```
rasaffie, line 275: * For `python >= 3.11`, `Action.Authorize` returns `"Action.Authorize"`.
* For `python < 3.11`, `Action.Authorize` returns `"Authorize"`.

poetry.lock
```diff
@@ -375,83 +393,115 @@ i18n = ["Babel (>=2.7)"]
 
 [[package]]
 name = "jsonschema"
-version = "4.17.3"
```
rasaffie, line 378: Dev dependency `asynctest` for `python@3.7` was setting version `4.17.3` of this library, so no `ImportError` is triggered when running tests or in development environment.
However, when installing `ocpp` in a project `asynctest` is not installed, so `jsonschema` is set to `4.19.1`, causing the `ImportError` in production environment.

tests/test_messages.py
rasaffie: Related to https://github.com/mobilityhouse/ocpp/issues/447

ocpp/messages.py
```diff
@@ -9,7 +9,7 @@
 from typing import Callable, Dict, Union
 
 from jsonschema import Draft4Validator
-from jsonschema import _validators as SchemaValidators
+from jsonschema import _keywords as SchemaValidators
```
OrangeTux, line 12: #479 already address the `ImportError` when using the version of jsonschema 4.19.0 and up. 
But I still want to inform you that this fix would break for everyone using a version older than 4.19.0.

OrangeTux:
Thanks for  your PR. It contains some useful changes. 

I've some feedback though:
1. Please provide one PR per issue. This PR fixes multiple issues (e.g. dropping support for Python 3.7, adding support for Python 3.11, fixing an `ImportError`for modern version of `jsonschema` etc). It's good practice to isolate 1 PR per issue. Among other things, that allows it revert a single fix easily, without reverting other fixes as well. 

2. This PR introduces the dependency [`StrEnum`](https://pypi.org/project/StrEnum/) to fix #447 . I'm conservative when adding dependencies to this library. A dependency always carries the risk of causing problems. For example, it's API could change, it's license could change and so on. In #447 I already proposed 2 fixes that don't rely on an extra dependency. I'm having a chat with my colleagues to discuss which solution mention in #447 we prefer.

ocpp/v20/schemas/BootNotificationRequest_v1p0.json
OrangeTux: Can you elaborate on why this change is required? What problem does this change solves?

poetry.lock
OrangeTux: I don't understand the relation between `asynctest` and `jsonschema`. Can you elaborate?
When dropping support for Python 3.7,  the dependency `asynctest` becomes obsolete and can be dropped.

pyproject.toml
```diff
@@ -52,4 +51,4 @@ profile = "black"
 
 [build-system]
 requires = ["poetry>=1.1.11"]
-build-backend = "poetry.masonry.api"
+build-backend = "poetry.core.masonry.api"
```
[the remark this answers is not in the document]
rasaffie: To be up to date with the latest documentation of Poetry, but I think both notations have the same result.

ocpp/v20/schemas/BootNotificationRequest_v1p0.json
OrangeTux: Interesting find, thanks for sharing. 

> When removing support for Python 3.7, some libraries versions are bumped.

That doesn't have to happen. After changing the `pyproject.toml` you can run `poetry lock --no-update`  to avoid updating any packages

OrangeTux: I was wrong about `poetry lock --no-update`.  Please ignore that comment.

I'm able to reproduce the issue shown in the screenshot using Python 3.10.

This assumes a clean clone from current master branch.

1. Modify `pyproject.toml` to update to set minimal python version to 3.8.
```diff
diff --git a/pyproject.toml b/pyproject.toml
index f48fbcd..052272a 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ classifiers = [
 ]

 [tool.poetry.dependencies]
-python = "^3.7"
+python = "^3.8"
 jsonschema = "^4.4.0"

 [tool.poetry.dev-dependencies]
 ```
 
 2. Run `poetry update` 
 3. Run `poetry run pytest`

Comment (OrangeTux) [reply ocpp/v20/schemas/BootNotificationRequest_v1p0.json:34]: I've created an issue to track this specific issue: #480.

I'm not sure what the best way to fix is to fix this error. We could modify all schemas. However, this project contains a lot of schemas. It's easy to forget to an incorrect reference. 

Do you know if it's possible to use `jsonschema` in a way it ignores these invalid references?

Comment (rasaffie) [reply ocpp/v20/schemas/BootNotificationRequest_v1p0.json:34]: Sorry, I don't 😞

**Comment by JoeriHermans** [comment]
Hi @rasaffie nice PR!

I believe you also have to modify the GitHub workflow, in order to be consistent. Changing
```
       version:
        - "3.7"
        - "3.8"
        - "3.9"
        - "3.10"
 ```
 in `.github/workflows/pull-request.yml` to
 ```
       version:
        - "3.8"
        - "3.9"
        - "3.10"
        - "3.11"
 ```
should do the trick I believe. Maybe it is a good idea to immediately check for Python 3.12, since it has been released? The changes should be minimal, I believe the only problem is a linting issue in `ocpp/exceptions.py`. In addition, there is also a small typo in the `help` description of the `Makefile`. It should be `tests` instead of `test`.

Jared-Newell-Mobility:
Replaced by PR:

https://github.com/mobilityhouse/ocpp/issues/564
https://github.com/mobilityhouse/ocpp/issues/565

So will close for now

PR #486, opened 2023-10-11
Update documents for 2.0.1 to lastest

Jared-Newell-Mobility:
Also remove 2.0 docs


OrangeTux (approved):

OrangeTux (changes requested):
It's cumbersome task to add this enums, so thanks for the PR! 

I didn't finish the review yet, since I found a few enums which are missing variants. That makes me wonder, based on which documents are the enums based?  Are they based on the latest errata?

Jared-Newell-Mobility:
@OrangeTux - for comment https://github.com/mobilityhouse/ocpp/pull/486#pullrequestreview-1679322180 - I'll reply under https://github.com/mobilityhouse/ocpp/pull/487 which I think this is what you are referring to

OrangeTux:
Yeah, your right. I posted my comment in the wrong PR. Let's keep the discussion at #487

OrangeTux (approved):

Issue #493, opened 2023-10-19
Improve error handling in ocpp/charge_point.py


Jared-Newell-Mobility:
see https://github.com/mobilityhouse/ocpp/pull/426

Table 7 of the Open Charge Point Protocol JSON 1.6, OCPP-J 1.6 Specification lists all error codes with a short description.

NotSupported - Requested Action is not known by receiver
NotImplemented - Requested Action is recognized but not supported by the receiver charge_point.py:175 can indicate both problems.
Trying to send a request with action "InvalidAction" triggers this code path. In this case, a NotSupportedError would be fitting.

However, the code path is also executed when the receiver receives a request for an action the receiver doesn't implement. For example, if a charger receives a "GetDiagnostics" request without supporting this feature. Now NotImplementedError would be suitable.

A while ago, we changed the error type from "NotImplementedErrortoNotSupportedError`. See https://github.com/mobilityhouse/ocpp/commit/c6e7e484f4cc263fb15c5f0802e70fe8eced1944

I'm hesitant to revert that change by approving this PR. Instead, some code should be include to distinguish between 2 scenarios I described and raise the appropriate error.

PR #494, opened 2023-10-19
Add NotImplemented exception

drc38:
See #426 for background discussion

### DIFF
--- a/ocpp/charge_point.py
+++ b/ocpp/charge_point.py
@@ -10,8 +10,6 @@ from typing import Dict, List, Union
 from ocpp.exceptions import NotImplementedError, NotSupportedError, OCPPError
 from ocpp.messages import Call, MessageType, unpack, validate_payload
 from ocpp.routing import create_route_map
-from ocpp.v16.enums import Action as v16_Action
-from ocpp.v201.enums import Action as v201_Action
 
 LOGGER = logging.getLogger("ocpp")
 
@@ -88,6 +86,9 @@ def _raise_keyerror(action, version):
     not implemented by the server/client and raises
     the appropriate error.
     """
+
+    from ocpp.v16.enums import Action as v16_Action
+    from ocpp.v201.enums import Action as v201_Action
 
     if version == "1.6":
         if hasattr(v16_Action, action):
--- a/ocpp/charge_point.py
+++ b/ocpp/charge_point.py
@@ -189,8 +189,8 @@ class ChargePoint:
             else:
                 raise NotSupportedError(
                     details={
-                        "cause":
-                        f"{msg.action} not supported by OCPP{self._ocpp_version}."
+                        "cause": f"{msg.action} not supported by OCPP"
+                        "{self._ocpp_version}."
                     }
                 )
 
@@ -220,8 +220,8 @@ class ChargePoint:
             else:
                 raise NotSupportedError(
                     details={
-                        "cause":
-                        f"{msg.action} not supported by OCPP{self._ocpp_version}."
+                        "cause": f"{msg.action} not supported by OCPP"
+                        "{self._ocpp_version}."
                     }
                 )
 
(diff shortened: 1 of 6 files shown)


Jared-Newell-Mobility:
Linked issue - https://github.com/mobilityhouse/ocpp/issues/493

OrangeTux:
You could move the imports of the `Action` enums to appear after the `ChargePoint` class. Not sure if that's the nicest solution, though.

ocpp/charge_point.py
```diff
@@ -165,17 +167,32 @@ async def _handle_call(self, msg):
...
+                    raise NotImplementedError(
+                        details={"cause": f"No handler for {msg.action} registered."}
+                    )
+            else:
+                raise NotSupportedError(
+                    details={
+                        "cause": f"{msg.action} not supported by OCPP"
+                        "{self._ocpp_version}."
+                    }
+                )
```
OrangeTux, line 195: This code is repeated twice. I suggest to isolate it to a function. 

Since the logic doesn't really need access to self, I don't think a method bound to the `ChargePoint` class is needed. Instead, just create a "private" function in the module scope.

OrangeTux (changes requested):
Thanks for the PR! 

Can you cover the changes by unit tests?

ocpp/charge_point.py
drc38: Thanks, done

drc38:
@OrangeTux, not sure why the mock is not being called to test an invalid action. Appreciate your assistance.

drc38:
I notice `messages.py` can raise a `NotImplementedError`, imo this is incorrect and should be a `ValidationError`

drc38:
Suggest amending to the following:
```
except OSError:
        raise ValidationError(
            details={"cause": f"Validation schema not found for action: {message.action}"}
        )
except json.JSONDecodeError:
      raise ValidationError(
            details={"cause": f"Error in decoding JSON validation schema for action: {message.action}"}
        )
```

tests/v16/conftest.py
```diff
@@ -17,6 +17,11 @@ def heartbeat_call():
     return Call(unique_id=1, action=Action.Heartbeat, payload={}).to_json()
 
 
+@pytest.fixture
+def notsupported_call():
```
OrangeTux, line 21: Please use snake_case consistently, so write `not_supported_call`.

ocpp/charge_point.py
```diff
@@ -79,6 +79,39 @@ def remove_nones(data: Union[List, Dict]) -> Union[List, Dict]:
     return data
 
 
+def _raise_keyerror(action, version):
```
OrangeTux, line 82: This should be `_raise_key_error`.

OrangeTux (changes requested):
Almost there!

OrangeTux:
@drc38 Are you referring to this? https://github.com/mobilityhouse/ocpp/blob/e452f5c3505495c033f0f8f9ad1ff05d98768314/ocpp/messages.py#L220-L223

When a schema doesn't exists, an `OSError` is raised. This mechanism is (mis)used to identify calls for non existing Actions. I think a `NotSupportedError` should be raised in that case.

If a file is found, but it can't be parsed as JSON a `json.JSONDecoder` error is raised. I agree with you that this error should be mapped to a `ValidationError`.

Let's fix that in a seperate PR.

drc38:
> @drc38 Are you referring to this?
> 
> https://github.com/mobilityhouse/ocpp/blob/e452f5c3505495c033f0f8f9ad1ff05d98768314/ocpp/messages.py#L220-L223
> 
> When a schema doesn't exists, an `OSError` is raised. This mechanism is (mis)used to identify calls for non existing Actions. I think a `NotSupportedError` should be raised in that case.
> 
> If a file is found, but it can't be parsed as JSON a `json.JSONDecoder` error is raised. I agree with you that this error should be mapped to a `ValidationError`.
> 
> Let's fix that in a seperate PR.

Ok. Note with this PR the `OSError` should only be raised on edge cases eg if the json schema file is corrupt. `NotImplementedError` and `NotSupportedError` errors will be raised prior to the json validation occurring, taking care of when the schema doesn't exist. Or have I missed something?

OrangeTux (approved):
Thanks!

PR #504, opened 2023-11-02
fix: add missing tech_info attribute to v2.0.1 EventDataType

LokiHokie:
The V2.0.1 datatype `EventDataType` is missing an optional string attribute for techInfo.


OrangeTux (approved):
Thanks!

PR #506, opened 2023-11-03
Amend validation error raised

drc38:
With the incorporation of #494 an OSError should only be raised on edge cases eg if the json schema file is corrupt. NotImplementedError and NotSupportedError errors will be raised prior to the json validation occurring, taking care of when the schema doesn't exist.

### DIFF
--- a/ocpp/messages.py
+++ b/ocpp/messages.py
@@ -13,7 +13,6 @@ from jsonschema.exceptions import ValidationError as SchemaValidationError
 
 from ocpp.exceptions import (
     FormatViolationError,
-    NotImplementedError,
     OCPPError,
     PropertyConstraintViolationError,
     ProtocolError,
@@ -217,10 +216,14 @@ def validate_payload(message: Union[Call, CallResult], ocpp_version: str) -> Non
             validator = get_validator(
                 message.message_type_id, message.action, ocpp_version
             )
-    except (OSError, json.JSONDecodeError):
-        raise NotImplementedError(
-            details={"cause": f"Failed to validate action: {message.action}"}
+    except OSError:
+        raise ValidationError(
+            details={"cause": f"JSON validation schema not found for action: {message.action}"}
         )
+    except json.JSONDecodeError:
+       raise ValidationError(
+            details={"cause": f"Error in decoding JSON validation schema for action: {message.action}"}
+       )
 
     try:
         validator.validate(message.payload)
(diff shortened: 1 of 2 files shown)


OrangeTux (approved):

ocpp/messages.py
```diff
@@ -217,9 +216,13 @@ def validate_payload(message: Union[Call, CallResult], ocpp_version: str) -> Non
...
-    except (OSError, json.JSONDecodeError):
-        raise NotImplementedError(
-            details={"cause": f"Failed to validate action: {message.action}"}
+    except OSError:
+        raise ValidationError(
+            f"JSON validation schema not found for action: {message.action}"
+        )
+    except json.JSONDecodeError:
+        raise ValidationError(
+            f"Error in decoding JSON validation schema for action: {message.action}"
```
OrangeTux, line 225: Only a `json.JSONDecodeError` should be mapped to a `ValidationError`. An `OSError` means that no schema with a given name was found. That means user is using a action that doesn't exists. Therefore, a `NotSupportedError` is applicable.  See my comment here: https://github.com/mobilityhouse/ocpp/pull/494#issuecomment-1786702680

OrangeTux (changes requested):

ocpp/messages.py
drc38: With #494 a `FileNotFoundError` should not be raised by the `OSError` exception, as before it gets to this part of the code a `NotSupportedError` will be raised by`_raise_key_error`. So an `OSError` should only be raised for one of these other reasons https://docs.python.org/3/library/exceptions.html#os-exceptions

Jared-Newell-Mobility:
If possible could you please add to this PR an entry in ChangeLog.md?

jainmohit2001:
@drc38 Should we take this up?
Can you please share e brief about the changes done in this PR. Sorry, I am out of loop for this one.
Also, please remove the changes in CHANGELOG.md. I believe we will handle that while creating a new release.

drc38:
Hi @jainmohit2001 as per discussion above these two errors should only be raised on edge cases if the validation has failed. Prior to this part of the code being executed a `NotSupportedError` if the ocpp spec does not include the message sent or `NotImplementedError` if server/client using this library has not coded to process message will be raised. Hence if neither is raised and a `json.JSONDecodeError` occurs it should be raised as a `ValidationError`. If an `OSError` occurs it should also be raised as a `ValidationError` because it indicates for some reason the filesystem and .json files cannot be read (seems very unlikely to occur).

My view is it should be merged.

PR #514, opened 2023-11-07
V2.0.1 unit of measure update docstring to allow use for variable characteristics

### LINKED ISSUE #510: v2.0.1 UnitOfMeasureType - update docstring to allow use for variableCharacteristics
Enums - UnitOfMeasureType 

1. is missing B for bytes

doc string can be extended as it is also used for component/variables - specifically unit in variableCharacteristics

Jared-Newell-Mobility:
Fixes #510

### DIFF
--- a/ocpp/v16/enums.py
+++ b/ocpp/v16/enums.py
@@ -753,8 +753,7 @@ class UnitOfMeasure(str, Enum):
     """
     Allowable values of the optional "unit" field of a Value element, as used
     in MeterValues.req and StopTransaction.req messages. Default value of
-    "unit" is always "Wh". Also used in 2.0.1 component/variables -
-    specifically the unit in variableCharacteristics.
+    "unit" is always "Wh".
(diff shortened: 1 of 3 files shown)


ocpp/v16/enums.py
```diff
@@ -773,6 +774,7 @@ class UnitOfMeasure(str, Enum):
     k = "K"
     percent = "Percent"
     hertz = "Hertz"
+    b = "Bytes"
```
OrangeTux, line 777: This change is done to the v16 enum. It should be applied to the v201 enum, right?

OrangeTux (changes requested):

ocpp/v16/enums.py
Jared-Newell-Mobility: Yes, so it is only an addition now to the doc string

OrangeTux (approved):

Issue #431, opened 2023-03-17
Attributes with 'v2x' are serialized as 'V2x', but should be serialized as 'V2X'.


OrangeTux:
https://github.com/mobilityhouse/ocpp/blob/db25fbd6c693e83f280660e58c683228ae794e0a/ocpp/v201/extensions/v2x/datatypes.py#L85-L86

Serializing objects with attributes that have 'v2x' in the name, are not serialized correctly. E.g. `V2XChargeParametersType.ev_max_v2x_energy_request` is serialized as "evMaxV2xEnergyRequest", but it should be "evMinV2**X**EnergyRequest". This problem happens also to other objects.

Jared-Newell-Mobility:
Currently:

`@dataclass
class Bidirectional:
    ev_min_v2x_energy_request: Optional[str] = None


def test_serialise():

    expected = {'evMinV2XEnergyRequest': None}

    assert snake_to_camel_case(asdict(Bidirectional())) == expected`

{'evMinV2xEnergyRequest': None} != {'evMinV2XEnergyRequest': None)

Jared-Newell-Mobility:
PR https://github.com/mobilityhouse/ocpp/pull/522 is prepared for inclusion in v1.0.0

PR #522, opened 2023-11-08
Attributes with 'v2x' are serialized as 'V2x', but should be serialized as 'V2X'

Jared-Newell-Mobility:


Jared-Newell-Mobility:
This is the example you gave in https://github.com/mobilityhouse/ocpp/issues/431

OrangeTux:
Can you add a test to cover the change?

OrangeTux:
> This is the example you gave in https://github.com/mobilityhouse/ocpp/issues/431

Lol, I must have forgotten this.

tmh-grunwald-markus (approved):
The handling of `V2X` is ok, but to avoid confusion I would handle `SoC` at the start of a word in the same way (which is not the point of THIS PR).

OrangeTux (approved):
LGTM

Issue #523, opened 2023-11-08
The serialisation of soc to SoC should not occur in camel case if it is existing at the beginning of a field


Jared-Newell-Mobility:
https://github.com/mobilityhouse/ocpp/blob/7e14b1664588e9f732f21fc5ff08369693385f35/ocpp/charge_point.py#L45-L60

consider this fictional field:

```
@dataclass
class Bidirectional:
    soc_battery: Optional[str] = None
```

when serialised;

results in SoCBattery when it should be socBattery

`@dataclass
class Bidirectional:
    soc_battery: Optional[str] = None

def test_serialise():

    expected = {'socBattery': None}

    assert snake_to_camel_case(asdict(Bidirectional())) == expected`

test_v2x serialisation.py:38 (test_serialise)
{'SoCBattery': None} != {'socBattery': None}

Jared-Newell-Mobility:
Merged so will close

PR #527, opened 2023-11-13
The serialisation of soc to SoC should not occur in camel case if it is existing at the beginning of a field

Jared-Newell-Mobility:


OrangeTux (commented):
Do you've an example of an input that which casing is changed incorrectly?

Jared-Newell-Mobility:
Just checking there is an example here - https://github.com/mobilityhouse/ocpp/issues/523 ?

OrangeTux (changes requested):
Can you update the [tests](https://github.com/mobilityhouse/ocpp/blob/c6129db91d00d721b51cca170a4a8f38225c9d85/tests/test_charge_point.py#L63-L72) to cover this change?

OrangeTux (approved):

PR #537, opened 2023-11-30
fix: DataTransfer data types

mdwcrft:
In 2.0.1 both DataTransferRequest and DataTransferResponse have a field called `data` which is defined as `AnyType` so I believe these should both by of type `Any` here

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Change log
 
 
+- [#537] (https://github.com/mobilityhouse/ocpp/pull/537) Fix DataTransfer data types 
 - [#531] (https://github.com/mobilityhouse/ocpp/pull/531) Add 1.6 security extension datatypes
 =======
 - [#528](https://github.com/mobilityhouse/ocpp/issues/528) v2.0.1 CertificateHashDataChainType childCertificateHashData requires the default of None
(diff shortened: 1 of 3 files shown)


Jared-Newell-Mobility (approved):

Jared-Newell-Mobility (changes requested):
Nice catch, could you also add an entry in the ChangeLog.md?

Jared-Newell-Mobility (approved):

OrangeTux (approved):

malsyned:
My read on the spec for this is different. The Field Type for `data` in both `DataTransferRequest` and `DataTransferResponse` is indeed listed as `anyType`. However, in **2.1.3. Primitive Datatypes**, `AnyType` is defined as "**Text**, data without specified length or format" (emphasis mine). So I think the correct Python data type for this should be `str`, not `Any`.

This is a pretty confusing part of the spec, though, I could certainly be wrong. It looks like it has a mistake in it as well, because the capitalization isn't consistent between the Field Types `anyType` and the Primitive Datatype name `AnyType`. Maybe this part of the spec needs some TLC in the next revision.

OrangeTux:
@malsyned Good catch! You're right. I think we should revert this change.

mdwcrft:
> My read on the spec for this is different. The Field Type for `data` in both `DataTransferRequest` and `DataTransferResponse` is indeed listed as `anyType`. However, in **2.1.3. Primitive Datatypes**, `AnyType` is defined as "**Text**, data without specified length or format" (emphasis mine). So I think the correct Python data type for this should be `str`, not `Any`.

Interesting, I read "Text, data" to mean "Text OR data" - agreed it's not clear though I think @OrangeTux will ask the OCA for a clarification

Issue #544, opened 2023-12-14
Pass the `Call.unique_id` to the `on` and `after` routing handlers.


wafa-yah:
At the level of [ChargePoint.handle_call](https://github.com/mobilityhouse/ocpp/blob/master/ocpp/charge_point.py#L195), currently the library only passes the message  payload to the `_on_action`, and `_after_action` handlers.
For some features like logging the messages being treated by those handlers, it would be nice to pass the `Call.unique_id` as well.

This will be a **breaking change** though as all the handlers defined by the library clients should always define `call_unique_id` in the kwargs, hence the label `For inclusion in release 1.0.0`

Jared-Newell-Mobility:
PR https://github.com/mobilityhouse/ocpp/pull/545 merged into master so will now close

PR #545, opened 2023-12-14
Add call unique id to handler kwargs

wafa-yah:
At the level of [ChargePoint.handle_call](https://github.com/mobilityhouse/ocpp/blob/master/ocpp/charge_point.py#L195), currently the library only passes the message payload to the _on_action, and _after_action handlers.
For some features like logging the messages being treated by those handlers, it would be nice to pass the Call.unique_id as well.

~~:warning: This will be a **breaking change** though as all the handlers defined by the library clients should always define call_unique_id in the kwargs, hence the label For inclusion in release 1.0.0~~ => Not a breaking change anymore after introducing the signature check.

Issue Link: https://github.com/mobilityhouse/ocpp/issues/544


OrangeTux:
I share your need for allowing users to access the request id. Currently, this library is lacking at that point.

 But I'm hesitant about this PR. It is a breaking change to the main API: the handler.  So potentially, it breaks a lot of code for users of this library.

I'm thinking out loud here and considering some alternative implementations:

1. Would it be possible to inspect the interface of a handler and only pass the unique id to the handlers when an argument `unique_id` is part of the signature?
2. Alternatively, in Flask, users have access to [the request context](https://flask.palletsprojects.com/en/3.0.x/reqcontext/). Maybe we can do something similar using[ ContextVars.](https://docs.python.org/3/library/contextvars.html).


I'm going to be unavailable during the Holidays, so I might respond in a few weeks.

wafa-yah:
> I share your need for allowing users to access the request id. Currently, this library is lacking at that point.
> 
> But I'm hesitant about this PR. It is a breaking change to the main API: the handler. So potentially, it breaks a lot of code for users of this library.
> 
> I'm thinking out loud here and considering some alternative implementations:
> 
>     1. Would it be possible to inspect the interface of a handler and only pass the unique id to the handlers when an argument `unique_id` is part of the signature?
> 
>     2. Alternatively, in Flask, users have access to [the request context](https://flask.palletsprojects.com/en/3.0.x/reqcontext/). Maybe we can do something similar using[ ContextVars.](https://docs.python.org/3/library/contextvars.html).
> 
> 
> I'm going to be unavailable during the Holidays, so I might respond in a few weeks.

@OrangeTux makes sense.
I applied the suggestion of checking the signature, should not be a breaking change now.

Jared-Newell-Mobility (dismissed):
Thanks for the clear definition in the tests

PR #546, opened 2023-12-17
Fix bug for ZJ Beny charger

thushan1994:
For ZJ Beny chargers - incoming data includes a misformatted array for sampleValues - [,{obj}, {obj}]. This fix removes the first comma. There would be no data requiring a comma immediately after a [ sign so should not cause interference otherwise.


Jared-Newell-Mobility:
Is not the problem here with ZJ Beny charger's implementation?

Issue #255, opened 2021-09-28
ocpp.v201.datatypes fail validation when optional values are not provided


MacDue:
For example, if you use class `ComponentType` (from `ocpp.v201.datatypes`)

Which looks like this:
```py
@dataclass
class ComponentType:
    name: str
    instance: Optional[str] = None
    evse: Optional[EVSEType] = None
```

and create the instance:
```py 
ComponentType(name="EVSE")
```

If you use that instance somewhere within a payload, you'll get a `ocpp.exceptions.ValidationError` as `ComponentType` will be serialized as:
```json
{ "name": "EVSE", "instance": None, "evse": None }
```

rather than the correct:
```json
{ "name": "EVSE" }
```

 I only used `ComponentType` as an example here, the same thing happens with all the data types with ``Optional[<type>] = None``

I don't think there's any way to fix this by changing the data types (a `dataclass` will always include all the fields), I think there probably needs to be another step that removes the optional fields that are `None` when the dataclass is converted to JSON.

---

P.s. 
The `SampledValueType` is missing the `measurand` field.

proelke:
Thats correct. At present the solution would be something like this:

`
from dataclasses import asdict
from ocpp. charge_point import remove_nones
from ocpp.v201.datatypes import ComponentType

component_type = ComponentType(name='EVSE')
component_type = remove_nones(asdict(component_type))
`

Admittedly not very elegant. The proper fix is to adjust remove_nones to handle this. I was working on a PR to do this but got a little sidetracked with other things.

MacDue:
I think it should probably be able to be done without requiring you to call any additional functions on the class/instance,
for example, in Python 3.8+ you could make an optional-aware asdict that uses the typing information within the dataclass to decide to remove the nones, and call that here: https://github.com/mobilityhouse/ocpp/blob/master/ocpp/messages.py#L232

```py
import typing
from dataclasses import asdict

# Note: get_origin() and get_args() are Python 3.8+ 
# (though there's ways to do the same in older versions)
def is_optional(field):
  # typing.Optional[x] is an alias for typing.Union[x, None]
  return (typing.get_origin(field) is typing.Union
          and type(None) in typing.get_args(field))

# Proof of concept (would be a little more complex for nested dataclasses types) 
def optional_aware_asdict(dataclass_instance):
  result = asdict(dataclass_instance)
  # Only remove 'None' values that are marked as optional in the dataclass
[...]

PR #547, opened 2023-12-18
Handle recursively serializing a dataclasses as a dictionary.

Jared-Newell-Mobility:
@OrangeTux this may have been missed so I created a PR - see https://github.com/mobilityhouse/ocpp/issues/255

Dependent on https://github.com/mobilityhouse/ocpp/pull/475 - which removes support for python 3.7

### DIFF
--- a/tests/test_charge_point.py
+++ b/tests/test_charge_point.py
@@ -84,7 +84,7 @@ def test_camel_to_snake_case(test_input, expected):
     [
         ({"transaction_id": "74563478"}, {"transactionId": "74563478"}),
         ({"full_soc": 100}, {"fullSoC": 100}),
-        ({"soc_limit_reached": 200}, {"SOCLimitReached": 200}),
+        ({"soc_limit_reached": 200}, {"SoCLimitReached": 200}),
         ({"ev_min_v2x_energy_request": 200}, {"evMinV2XEnergyRequest": 200}),
         ({"v2x_charging_ctrlr": 200}, {"v2xChargingCtrlr": 200}),
         ({"web_socket_ping_interval": 200}, {"webSocketPingInterval": 200}),


ocpp/charge_point.py
```diff
@@ -69,6 +69,85 @@ def snake_to_camel_case(data):
...
+            schema: str,
+            host: str,
+            post: Optional[str],
+            query: Union[None, str]
+
+    """
+    return get_origin(field.type) is Union and type(None) in get_args(field.type)
+
+
+def serialize_as_dict(dataclass, remove_empty_optional_fields: bool = True):
```
OrangeTux, line 94: I think the solution can be simplified.

The root of issue #255 is that `dataclasses.asdict()` only serializes the top level dataclass as a dictionary. Any inner dataclasses are not serializes as a dict. 

The body of this function fixes that: recursively serializing a dataclasses as a dictionary.

But this function does more:
1. remove any optional fields containing `None`.
2. change casing to snake_case.

I think we can remove the latter to features of this function.

And, in lines [318](https://github.com/mobilityhouse/ocpp/blob/71f57ea18a748743aaf1ec185525f17fddd2fc2b/ocpp/charge_point.py#L318) and [378](https://github.com/mobilityhouse/ocpp/blob/71f57ea18a748743aaf1ec185525f17fddd2fc2b/ocpp/charge_point.py#L373),  replace the call to `dataclass.asdict()` with `serialize_as_dict()`.

OrangeTux (changes requested):

pyproject.toml
```diff
@@ -33,6 +33,7 @@ classifiers = [
 [tool.poetry.dependencies]
 python = "^3.7"
 jsonschema = "^4.4.0"
+typing-extensions = "^4.0.0"
```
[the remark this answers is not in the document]
Jared-Newell-Mobility: Sounds good - Corrected with commit f59f3460e78b2b917f22d6a4f16369687e7a7ed0

ocpp/charge_point.py
```diff
@@ -69,6 +69,85 @@ def snake_to_camel_case(data):
...
+    """Serialize the given `dataclass` as a `dict` recursively.
+
+
+    @dataclass
+    class StatusInfoType:
+        reason_code: str
+        additional_info: Optional[str] = None
+
+    with_additional_info = StatusInfoType(reason="Unknown",
+    additional_info="More details")
```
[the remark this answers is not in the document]
Jared-Newell-Mobility: Done - satisfied with b3c62d3671f5541ac49ca15f123540ee4d060f27

ocpp/charge_point.py
Jared-Newell-Mobility: asdict() corrected in commit d1609599424af910fd596081afaafdc8fcef546d

Jared-Newell-Mobility: Removed the more with commit 2834dbefd92f9e072df5cdf81a163b96daef8a07

OrangeTux (approved):

Issue #579, opened 2024-01-17
v2.0.1 Action enums corrected


Jared-Newell-Mobility:
Quite a few OCPP v2.0.1 enums have memebers that don't follow Python naming guide lines. 
For example:

```python
 BootNotification = "BootNotification"
   ...
```

For a long time, these variants have been deprecated in favor of members following Python naming guide lines:

```python
boot_notification = "BootNotification"
   ...
```

This issue tracks the work to finally remove the deprecated, unpythonic members.
Users should upgrade use the Pythonic members instead.

Jared-Newell-Mobility:
Merged so will close

PR #580, opened 2024-01-17
v2.0.1 Action enums corrected

Jared-Newell-Mobility:
See issue https://github.com/mobilityhouse/ocpp/issues/579


Jared-Newell-Mobility:
I completely agree with you here, I will modify this PR and see if we incorporate the warning

Jared-Newell-Mobility:
I have added the previous members back into the Action and in their current order, the result is as follows:

- Action("BootNotification") results in 
<Action.BootNotification: 'BootNotification'>
- Action.BootNotification.name  results in 
'BootNotification'
- Action.BootNotification.value  results in 
'BootNotification'
- Action.BootNotification  results in 
<Action.BootNotification: 'BootNotification'>
- Action.boot_notification  results in 
<Action.BootNotification: 'BootNotification'>
- Action.boot_notification.name  results in 
'BootNotification'
- Action.boot_notification.value  results in 
'BootNotification'

The problem we have here is that the name of the members changes and the value remains the same. Due to this the member is never missing. 

So after looking into this using  __missing__ we can only:
- catch a member only when in it missing but in our case it is replaced
- if we could make use of it, it will works in this case when accessing via Action("BootNotification") but not Action.BootNotification.name (as this will no longer exist after removal)

So to remove these into the future becomes quite a conundrum ... however, the current change could be a logical step?

OrangeTux:
Not sure if I follow you. 

I did some tests and figured out that `_missing_()` is only called when you instantiate an enum like `Action("BootNotification")`. 

Normal lookups, like `Action.BootNotification`, won't call `_missing_()` and therefore won't print a warning.

I fixed that with this fix.

```python
from warnings import warn
from enum import StrEnum

class Action(StrEnum):
    boot_notification = "BootNotification"

    def __init__(self, *args, **kwargs):
        if self.name == 'BootNotification':
            warn("Action.BootNotification is deprecated and will be removed in the future. Please use Action.boot_notification.")

        return super().__init__(*args, **kwargs)


print(Action.BootNotification)
print(Action.BootNotification.name)
print(Action.BootNotification.value)
print(Action.boot_notification)
print(Action.boot_notification.name)
print(Action.boot_notification.value)
print(Action['boot_notification'])
print(Action("BootNotification"))
```

It prints:

```
/tmp/dep.py:10: UserWarning: Action.BootNotification is deprecated and will be removed in the future. Please use Action.boot_notification.
  warn("Action.BootNotification is deprecated and will be removed in the future. Please use Action.boot_notification.")
```

Jared-Newell-Mobility:
I tried the approach with __init__ but the warning is only happens once during initialisation of the enum and doesn't appear during its use: 

```
PyDev console: starting.
Python 3.11.7 (main, Dec  8 2023, 18:56:58) [GCC 11.4.0] on linux
from test_enums import Action
/home/jared/PycharmProjects/test_project/test_enums.py:11: UserWarning: Action.BootNotification is deprecated and will be removed in the future. Please use Action.boot_notification.
  warn(

Action.BootNotification
<Action.BootNotification: 'BootNotification'>
Action.boot_notification
<Action.BootNotification: 'BootNotification'>
Action.boot_notification.value
'BootNotification'
Action.boot_notification.name
'BootNotification'
```

So I attempted to modify this some more and got this:

```
from warnings import warn
from enum import StrEnum

class Action(StrEnum):

    def __init__(self):
        if self._name_ == "BootNotification":
            self._value_ = self._name_
        if self._name_ == "boot_notification":
            self._value_ = self._name_.title().replace("_","")

    BootNotification = ()
    boot_notification = ()

    @classmethod
    def _missing_(cls, value):
        if value == "BootNotification":
            warn(
                "Action.BootNotification is deprecated and will be removed in the future. Please use Action.boot_notification.")
            return cls.BootNotification
```
which results in:
```
from test_enums import Action

Action("BootNotification")
<Action.BootNotification: 'BootNotification'>
/home/jared/PycharmProjects/test_project/test_enums.py:18: UserWarning: Action.BootNotification is deprecated and will be removed in the future. Please use Action.boot_notification.
  warn(
Action.BootNotification.name
'BootNotification'
Action.BootNotification.value
'BootNotification'
Action.boot_notification.name
'BootNotification'
Action.boot_notification.value
'BootNotification'
Action.BootNotification
<Action.BootNotification: 'BootNotification'>
Action.boot_notification
<Action.BootNotification: 'BootNotification'>
```

Although this solves some of the problem, I couldn't get the new enum boot_notification to function correctly. I have identified that the order of members in the enums looks to determine this behaviour ....

OrangeTux:
Good catch. I didn't realize the warning is show, even if you don't use any deprecated variants.

I'm afraid we can't use the builtin warning to warn the users.

OrangeTux (changes requested):
I realized that the enums of v16 are also incorrect. Can you update those as well?

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 - [#564](https://github.com/mobilityhouse/ocpp/issues/564) Add support For Python 3.11 and 3.12
 - [#583](https://github.com/mobilityhouse/ocpp/issues/583) OCPP v1.6/v2.0.1 deprecate dataclasses from calls and call results with the suffix 'Payload'
 
+## BREAKING ##
```
OrangeTux, line 5: These changes are not yet breaking, I suggest to change the title to "Deprecated".

ocpp/v201/enums.py
```diff
@@ -11,6 +13,13 @@ class StrEnum(str, Enum):  # pragma: no cover
 class Action(StrEnum):
     """An Action is a required part of a Call message."""
 
+    def __init__(self, *args, **kwargs):
+        warn(
+            "Action enum contains deprecated members and will be removed in "
+            "the next major release, please use snake case members."
+        )
```
OrangeTux, line 20: Can you apply the comment I made [here](https://github.com/mobilityhouse/ocpp/pull/600#discussion_r1484154853) to this PR?

OrangeTux (changes requested):

OrangeTux:
You can forget my earlier comment to include the v16 action. I see that the v16 action in #600 .

ocpp/v201/enums.py
Jared-Newell-Mobility: Satisfied with commit [50e4a95](https://github.com/mobilityhouse/ocpp/pull/580/commits/50e4a955cb6418e703fd5e0ce52f23d94a9b714c)

CHANGELOG.md
Jared-Newell-Mobility: Satisfied with commit https://github.com/mobilityhouse/ocpp/pull/580/commits/acba23f4678c62d5b171f6eaf2d1ea5924449299

OrangeTux (approved):

PR #582, opened 2024-01-22
Update GetConfigurationResponse.json

putongyong:
Dear project owner,
Actually we have found that in the real world case, certain provider, for example Scame from Italy, configure the charger in such a way that the value of configuration could be a string, a number or a boolean. For example, the hearbeat interval or metervalues interval is of a type of integer. 

The same as change configuration, we sincerely proposed this change because otherwise the response that we get from the charger will trigger some error. Thanks

Yong


putongyong:
Thanks a lot, I called the vendor and they provided a new firmware that solved the problem.

PR #584, opened 2024-01-25
OCPP v1.6/v2.0.1 deprecate dataclasses from calls and call results with the suffix 'Payload'

### LINKED ISSUE #583: Remove suffex 'Payload' from all calls and call results
All the calls and call result types are suffixed with the term 'Payload'.
This term makes lengthy type names even more verbose and should be removed.

This change will break every application relying on this library. So I suggest we deprecate these old style types and urge users to use the new types. In a v2 of this library we can remove them.

Jared-Newell-Mobility:
Fixes #583

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Change log
 
+- [#583](https://github.com/mobilityhouse/ocpp/issues/583) OCPP v1.6/v2.0.1 deprecate dataclasses from calls and call results with the suffix 'Payload'
+
 ## 0.26.0 (2024-01-17)
 
 - [#544](https://github.com/mobilityhouse/ocpp/issues/544) ocpp/charge_point.py - Pass `Call.unique_id` to the `on` and `after` routing handlers.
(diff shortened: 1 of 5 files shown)


OrangeTux (changes requested):
I suggest to warn users using a `DeprecationWarning`. For example by doing something like this:


```python
import warnings
from dataclasses import dataclass

@dataclass
class CancelReservation:
    reservation_id: int


@dataclass
class CancelReservationPayload(CancelReservation):
    def __post_init__(self):
        warnings.warn("CancelReservationPayload is deprecated, use CancelReservation instead.")

CancelReservation(reservation_id=3)
CancelReservationPayload(reservation_id=3)
```

Creating a `CancelReservation()` doesn't have any different effect. Whereas creating a `CancelReservationPayload()` prints a warning to stdout:

```
/tmp/deprecation.py:12: UserWarning: CancelReservationPayload is deprecated, use CancelReservation instead
```

Initializing a `CancelReservation()` doesn't have any affect. Creating a `CancelReservationPayload()

OrangeTux (approved):

Issue #565, opened 2024-01-09
Drop support for Python 3.7


OrangeTux:
Python 3.7 reached End of Life. Therefore, we can stop testing against this version. 

We should:
1. [remove Python 3.7 from our workflow.](https://github.com/mobilityhouse/ocpp/blob/5f27521b5c7b83abfb61217036ae5776e1fb6998/.github/workflows/pull-request.yml#L14)
2. [raise the minimum Python supported Python version to 3.8](https://github.com/mobilityhouse/ocpp/blob/5f27521b5c7b83abfb61217036ae5776e1fb6998/pyproject.toml#L34)
3. [remove the dependency on `asynctest`](https://github.com/mobilityhouse/ocpp/blob/5f27521b5c7b83abfb61217036ae5776e1fb6998/pyproject.toml#L38-L39) That fixes #433 
4. Remove [Python 3.7 classifier](https://github.com/mobilityhouse/ocpp/blob/5f27521b5c7b83abfb61217036ae5776e1fb6998/pyproject.toml#L30) from pyproject.toml

MuhammadHasnain77:
I have remove the version 3.7 from workflow check it out
also can you explain the other two issue

OrangeTux:
Among other things, `pyproject.toml` lists all dependencies of this project. That includes [the Python version(s)](https://github.com/mobilityhouse/ocpp/blob/5f27521b5c7b83abfb61217036ae5776e1fb6998/pyproject.toml#L34) that are supported by this project. 

That file must be modified so minimum supported Python version is raised from Python 3.7 to Python 3.8.  Read more at Poetry's documentation about [dependency specification](https://python-poetry.org/docs/dependency-specification/).

`asynctest` is a dependency when using this project with Python 3.7. For all new Python version, this dependency is not required anymore. Since we drop support for Python 3.7, we can also remove the dependency `asynctest` completely. [Here](https://python-poetry.org/docs/dependency-specification/#python-restricted-dependencies) you can read more on restricting dependencies to specific Python version

MuhammadHasnain77:
Thanks for describing the issue I will  work on it.

Jared-Newell-Mobility:
Additional work to date can be found in https://github.com/mobilityhouse/ocpp/pull/475

MuhammadHasnain77:
Hi 
Can you see my PR did i correct all the issues?

Jared-Newell-Mobility:
[...]

PR #585, opened 2024-01-26
drop support for python 3.7

Jared-Newell-Mobility:
See issue https://github.com/mobilityhouse/ocpp/issues/565

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -11,7 +11,6 @@ jobs:
     strategy:
       matrix:
         version:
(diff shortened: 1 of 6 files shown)


OrangeTux (changes requested):
Python 3.7 must also be removed from the Github workflow. See https://github.com/mobilityhouse/ocpp/blob/5f27521b5c7b83abfb61217036ae5776e1fb6998/.github/workflows/pull-request.yml#L14

OrangeTux (approved):

Issue #590, opened 2024-02-05
URL does not get converted from snake_case responder_url to camelCase responderURL


Jared-Newell-Mobility:
In the 2.0.1 spec for AuthorizeRequest and GetCertificateStatusRequest there is a field responderURL that does not get converted from snake_case responder_url to camelCase responderURL. This PR adds handling of this edge case. in SetNetworkProfileRequest there is a field ocppCsmsUrl which is handled correctly but causes problems for a more general solution.

Let me know if the conditionals work for you.

See Original PR https://github.com/mobilityhouse/ocpp/pull/336

Jared-Newell-Mobility:
Merged so will close

PR #592, opened 2024-02-05
URL does not get converted from snake_case responder_url to camelCase responderURL

Jared-Newell-Mobility:
see issue https://github.com/mobilityhouse/ocpp/issues/590


ocpp/charge_point.py
```diff
@@ -54,6 +54,8 @@ def snake_to_camel_case(data):
         for key, value in data.items():
             key = key.replace("soc", "SoC")
             key = key.replace("_v2x", "V2X")
+            key = key.replace("_url", "URL")
```
proelke, line 57: The issue with this approach is another field, as I mentioned in my other PR. The ocppCsmsUrl field in a SetNetworkProfileRequest will fail due to this change. I would suggest adding test cases for any usage of URL in the spec.

OrangeTux: I believe these are all attributes that include `url`:

```
$ grep -i "url" ocpp/v201/{call,call_result,datatypes}.py
ocpp/v201/datatypes.py:    ocpp_csms_url: str
ocpp/v201/datatypes.py:    responder_url: str
```

OrangeTux (approved):

OrangeTux (changes requested):
The tests don't pass. Can you fix them?

Jared-Newell-Mobility:
This should now be complete - @proelke does this cover everything?

OrangeTux (approved):

PR #597, opened 2024-02-05
Revert "drop support for python 3.7"

Jared-Newell-Mobility:
Reverts mobilityhouse/ocpp#585

As this is breaking should be included in release 1.0.0

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -11,6 +11,7 @@ jobs:
     strategy:
       matrix:
         version:
+        - "3.7"
         - "3.8"
         - "3.9"
         - "3.10"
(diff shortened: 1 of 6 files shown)


OrangeTux:
What's  the reason Python 3.7 support had to be add again? Before dropping the support again in #598 ?

Jared-Newell-Mobility:
I'm a bit confused this was merged yesterday, but this should be part of a major release, are we moving now to 1.0.0?

OrangeTux:
Sure, Python 3.7 should be dropped. I'm as confused as you that Python 3.7 support was added again.

Jared-Newell-Mobility:
Lets then align on this later today

Issue #601, opened 2024-02-07
Fix case conversion for soc in non "State of Charge" context


Jared-Newell-Mobility:
Add all variant tests to support this

Jared-Newell-Mobility:
Merged so will close

PR #607, opened 2024-02-08
Fix case conversion for `soc` in non "State of Charge" context

Jared-Newell-Mobility:
see issue https://github.com/mobilityhouse/ocpp/issues/601

### DIFF
--- a/ocpp/charge_point.py
+++ b/ocpp/charge_point.py
@@ -53,7 +53,7 @@ def snake_to_camel_case(data):
     if isinstance(data, dict):
         camel_case_dict = {}
         for key, value in data.items():
-            key = key.replace("soc", "SOC").replace("_SOCket", "Socket")
+            key = key.replace("soc", "SoC").replace("_SoCket", "Socket")
             key = key.replace("_v2x", "V2X")
             components = key.split("_")
             key = components[0] + "".join(x[:1].upper() + x[1:] for x in components[1:])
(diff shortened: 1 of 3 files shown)


ocpp/charge_point.py
```diff
@@ -53,7 +53,7 @@ def snake_to_camel_case(data):
     if isinstance(data, dict):
         camel_case_dict = {}
         for key, value in data.items():
-            key = key.replace("soc", "SoC")
+            key = key.replace("soc", "SOC").replace("_SOCket", "Socket")
```
OrangeTux, line 56: `snake_to_camel_case()` converts terms from idiomatic Python to the casing that the OCPP spec uses.
In the context of "State of Charge", OCPP uses the term `SoC`, not `SOC`.

See for example in [NotifyEVChargingNeedsRequest.json
](https://github.com/mobilityhouse/ocpp/blob/ae02ede445ea56375505d21bd498a88fce937862/ocpp/v201/schemas/NotifyEVChargingNeedsRequest.json#L127).

Make sure that `soc` is mapped to `SoC` in the context of "State of Charge". But that it is not converted in other contexts, for example in the context of `socket`.

OrangeTux (changes requested):

ocpp/charge_point.py
Jared-Newell-Mobility: Ok, thanks for clear definition - I'll make the changes

Jared-Newell-Mobility: Satisfied in https://github.com/mobilityhouse/ocpp/pull/607/commits/1ad08b70a8d562d141559907f0e1560b8be0154e

OrangeTux (approved):

PR #615, opened 2024-02-27
Removing state that ties a ChargePoint instance to a specific connected CP

ben-garrett-ev-energy:
We want to be able to use the message parsing/creation and the `@on` annotations of the `ocpp` library, but we don't want to have `ChargePoint` objects tied to a specific connection - we want them to be stateless and able to handle messages for any connected CPs in a parallel architecture.

### DIFF
--- a/ocpp/async_message_handler.py
+++ b/ocpp/async_message_handler.py
@@ -1,0 +1,12 @@
+from abc import ABC
+from abc import abstractmethod
+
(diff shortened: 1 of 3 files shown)


ben-garrett-ev-energy:
Sorry, I'm new to forks and didn't want to raise a PR on the upstream repo!

Issue #631, opened 2024-05-08
Publishing package to pypi fails.


OrangeTux:
The [GA workflow](https://github.com/mobilityhouse/ocpp/actions/runs/7902772972/job/21569153956) to publish the package to pypi fails.

The issue is that the `README.rst` is not valid. 
One can check it's validity  using:

```python
$ python3 -m pip install build, twine
$ python3 -m build --sdist
$ twine check dist/*
Checking dist/ocpp-1.0.0.tar.gz: FAILED
ERROR    `long_description` has syntax errors in markup and would not be rendered on PyPI.
         line 62: Warning: Title underline too short.

         Charging Station Management System (CSMS) / Central System
         ~~~~~~~~~~~~~~
```
For more information, read
We should fix the issue and modify our GA workflow to prevent this issue from happening again.

For more information, read:
* https://packaging.python.org/en/latest/guides/making-a-pypi-friendly-readme/#validating-restructuredtext-markup
* https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#packaging-your-project

After the fixes, I suggest we create a v1.01 release

PR #636, opened 2024-05-21
Fix release to PYPI

OrangeTux:
This commit fixes the README.rst so that Pypi accepts it.

Fixes: #631


OrangeTux:
@OSkrk Can you also review this one? For unknown reasons I can't select you as reviewer.

Issue #647, opened 2024-06-12
Blocking call detected


drc38:
The ocpp integration (https://github.com/lbbrhzn/ocpp) using this library in Home Assistant 2024.06 has detected that the IO `open` file call in `messages.py` blocks the event loop.

Would it make sense in `charge_point.py` to replace calls to `validate_payload` with the equivalent run in the executor eg `await asyncio.get_event_loop().run_in_executor(...,validate_payload,...)`?

Or if you have suggestions on how it can be addressed another way that would be great.

Many thanks

drc38:
Confirmed the following example fixes the blocking error:
```
            await asyncio.get_event_loop().run_in_executor(
                None,
                validate_payload,
                msg,
                self._ocpp_version
            )
```
Let me know if you want me to submit a PR

koopee:
> Confirmed replacing the calls (eg the below example) fixes the blocking error:
> 
> ```
>             await asyncio.get_event_loop().run_in_executor(
>                 None,
>                 validate_payload,
>                 msg,
>                 self._ocpp_version
>             )
> ```
> 
> Let me know if you want me to submit a PR

I don't really know how the process works, but maybe you should submit PR...

jainmohit2001:
FIxed in https://github.com/mobilityhouse/ocpp/pull/652

astrand:
I am requesting a revert of this code, see https://github.com/mobilityhouse/ocpp/pull/676 . 

I think the original report should be clarified: Is really the open() system call a problem, or subsequent calls? It is a very strange limitation of not being able to use "open" in an asyncio application. For long running executions and heavy IO usage, it is possible to yield to event loop with asyncio.sleep(0).

PR #652, opened 2024-06-24
add executor to validate_payload

drc38:
Addresses issue #647

### DIFF
--- a/ocpp/charge_point.py
+++ b/ocpp/charge_point.py
@@ -296,10 +296,7 @@ class ChargePoint:
 
         if not handlers.get("_skip_schema_validation", False):
             await asyncio.get_event_loop().run_in_executor(
-                None,
-                validate_payload,
-                msg,
-                self._ocpp_version
+                None, validate_payload, msg, self._ocpp_version
             )
         # OCPP uses camelCase for the keys in the payload. It's more pythonic
         # to use snake_case for keyword arguments. Therefore the keys must be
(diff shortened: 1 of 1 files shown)


drc38:
> Hi @drc38, the changes look good. I believe the functionality will not break and it is backwards compatible as well.
> 
> Here's another thought: Instead of using `run_in_executor` to run the blocking `validate_payload` function call, what if we convert the `validate_payload` function into a pure async function with all I/O calls being non-blocking.

That would be result in quite a departure from the current structure where most functions are not async. Another approach is to use https://github.com/Tinche/aiofiles for the file open, assuming you're happy to import aiofiles as part of the project.

drc38:
> That would result in quite a departure from the current structure where most functions are not async. Another approach is to use https://github.com/Tinche/aiofiles for the file open, assuming you're happy to import aiofiles as part of the project.

Actually using aiofiles would still require a rewrite to async. I think the proposed approach is the best for now.

jainmohit2001:
LGTM.

proelke (approved):

drc38:
@proelke @jainmohit2001 Thank you both for continuing to maintain the library,  are we able to get a PyPi release that includes this merge please.

N3rdix:
Bumping this, can we support somehow to get a PyPi release shipped?

mdwcrft:
@N3rdix @drc38 discussion on a new release is on #675 - aiming for this weekend

PR #678, opened 2024-11-18
Disable threads if messages.ASYNC_VALIDATION = False

astrand:
See discussion at https://github.com/mobilityhouse/ocpp/pull/676


jainmohit2001:
Thanks for the great PR and in-depth discussion. Really appreciate the time you have spent to ensure backward compatibility.
Can you please add a test where we set the `ASYNC_VALIDATION` variable to `True`?

Issue #264, opened 2021-11-09
Passing the response from @on to the @after method


HugoJP1:
Hello,

I have the following scenario:

```python
    @on(Action.RequestStartTransaction)
    async def on_request_start_transaction(
        self,
        *args,
        **kwargs
    ) -> call_result.RequestStartTransactionPayload:

        response: call_result.RequestStartTransactionPayload = (
            await TransactionHandler.handle_on_request_start_transaction(**kwargs)
        )

        # at this point, the response can be accepted
        # or it can be rejected
        return response


    @after(Action.RequestStartTransaction)
    async def after_request_start_transaction(
        self,
        *args,
        **kwargs
    ) -> call_result.TransactionEventPayload:

        # at this point, I would like to run this method conditionally,
        # only if the response sent back to the CSMS was accepted
        # if it was rejected, I do not want to run this method
        return await TransactionHandler.handle_after_request_start_transaction(**kwargs)

```

Looking at `charge_point.py`, it looks like it is the payload from the initial call which is passed into the `@after` function. What would you think of also passing through the response of the call so that it can be used for further processing?

I have a workaround by using a temp variable, but thought it could be a nice feature of the library so posting here.

tropxy:
I think this is a good suggestion and the enhancement should not be too hard as we only need to pass the response to the handler here:
https://github.com/mobilityhouse/ocpp/blob/efc8f081d063bb43680fd8f2b8c861a71356d784/ocpp/charge_point.py#L221

OrangeTux:
The routing system of this library is not very sophisticated and could use some improvement.

In the current implementation, the `@on()` handler is executed and it's return value is send as response to the other peer. Next, the `@after()` handler is executed. The return value of the `@after()` handler is ignored. 

I'm afraid that above process might not be clear to all users of this library.  In fact, the code sample that @HugoJP1 seem to indicate that this is not clear for them, as the `@after()` contains a `return` statement. Although this return value doesn't do anything.  Don't get me wrong, that is a fault of the API, not of the users. 


Assume for now,  return value of the `@on()` handler is passed down to the `@after()` handler. Modifying the response and trying return it won't send the modified response to the other peer.  Here a code sample that shows invalid use of the proposed API:

``` python
class ChargePoint(cp)
    @on(Action.BootNotification)
    def on_boot_notification(self, charge_point_vendor: str, charge_point_model: str, **kwargs):
        return call_result.BootNotificationPayload(
[...]

PR #683, opened 2024-11-26
Feat/allow passing answer from @on to @after hook

OSkrk:
This PR addresses the feature discussed [here](https://github.com/mobilityhouse/ocpp/issues/264), allowing the response sent inside the @on to be passed to the @after hook for further processing (avoiding the use of temp variables). The feature enables the @after decorator to accept an `inject_response` parameter, which can be used to control whether the response from the `_on_action` handler is passed to the `_after_action` handler, and guarantees backwards compatibility.

@OrangeTux , @tropxy, @HugoJP1, @mdwcrft would like to know you guy's feedback on this.


tests/test_charge_point.py
```diff
@@ -472,3 +472,33 @@ def after_boot_notification(self, *args, **kwargs):
...
+        def on_boot_notification(self, **kwargs):
+            return BootNotificationResult(
+                current_time="2024-11-01T00:00:00Z",
+                interval=300,
+                status=RegistrationStatus.accepted,
+            )
+
+        @after(Action.BootNotification, inject_response=True)
+        def after_boot_notification(self, on_response, **kwargs):
+            # print args and kwargs to check if response is injected
```
OSkrk, line 495:
```suggestion
```

ocpp/charge_point.py
```diff
@@ -356,10 +356,25 @@ async def _handle_call(self, msg):
...
+                        **snake_case_payload, call_unique_id=msg.unique_id
+                    )
             else:
-                response = handler(**snake_case_payload)
+                if inject_response:
+                    response = handler(
+                        **snake_case_payload, on_response=response_payload
+                    )
+                else:
+                    response = handler(**snake_case_payload)
```
astrand, line 377: Not so "elegant" to have 4 calls to handler(). My suggestion is to use keyword arguments instead, like I did in https://github.com/mobilityhouse/ocpp/issues/264#issuecomment-1910237896 .

ocpp/charge_point.py
```diff
@@ -356,10 +356,25 @@ async def _handle_call(self, msg):
             call_unique_id_required = "call_unique_id" in handler_signature.parameters
             # call_unique_id should be passed as kwarg only if is defined explicitly
             # in the handler signature
+            inject_response = getattr(handler, "_inject_response", False)
```
astrand, line 359: Why check attribute on the handler, instead of checking signature parameters? This opens up for some corner cases, like if inject_response is True but after handler does not accept "on_response" kwarg. Also it is different from how call_unique_id is handled.

astrand (changes requested):
Thanks for working on this!

ocpp/charge_point.py
OSkrk: @astrand , thanks for your comment. Why would you set the flag to `True` to pass the reponse and do not accept it in the @after ?

ocpp/charge_point.py
OSkrk: This approach was suggested by @tropxy (one of the authors of this repo) in this [comment](https://github.com/mobilityhouse/ocpp/issues/264#issuecomment-964425685)

ocpp/routing.py
```diff
@@ -56,7 +56,7 @@ def inner(*args, **kwargs):
     return decorator
 
 
-def after(action):
+def after(action, inject_response=False):
```
tropxy, line 59: can you include notes in the doc strings in respect to the new argument and its purpose?

ocpp/charge_point.py
```diff
@@ -356,10 +356,25 @@ async def _handle_call(self, msg):
...
             if call_unique_id_required:
-                response = handler(**snake_case_payload, call_unique_id=msg.unique_id)
+                if inject_response:
+                    response = handler(
+                        **snake_case_payload,
+                        call_unique_id=msg.unique_id,
+                        on_response=response_payload,
```
tropxy, line 365: It would be better the term "call_response". The "on" prefix is given to functions/callbacks as an action will be performed via that callback function. Here we want to pass an argument..

ocpp/charge_point.py
tropxy: @OSkrk what @astrand is suggesting is something like this (right @astrand ?):

```python
if "call_unique_id" in handler_signature.parameters:
    snake_case_payload["call_unique_id"] = msg.unique_id
if  inject_response:
    snake_case_payload["on_response"] = response_payload

```

May you try that and see if it works?

Note: I have used your current nomenclature, but if you change the code then "on_response" will be "call_response"

tropxy (changes requested):

OrangeTux:
I don't have anything against this PR. I leave it up to the new maintainers to approve and merge it.

I'll have a call with the new maintainers this Friday and I'll bring this PR to the table.

ocpp/routing.py
OSkrk: added a docstring section to the after() decorator in ocpp/routing.py

ocpp/charge_point.py
OSkrk: Addressed

ocpp/charge_point.py
OSkrk: Addressed. The rename from on_response → call_response is done everywhere

ocpp/charge_point.py
OSkrk: Keeping the explicit inject_response flag as discussed (and per @tropxy / @OrangeTux's validation). Injecting call_response is consistent with how payload kwargs are already passed... you opt in with the flag and accept it in the handler (or via **kwargs). If someone sets the flag but doesn't accept the arg, they get an immediate, clear TypeError, so it fails fast and only on explicit opt-in.

astrand (approved):

mdwcrft (approved):

PR #693, opened 2024-12-09
ci: ensure the matching poetry version is used

jerome-benoit:
* We should not be updating dependencies in CI. Hence we should remove the `poetry update` command
* To install poetry, following the official documentation, we should use the following command: https://python-poetry.org/docs/#installing-with-pipx
```shell
pipx install poetry
```


jainmohit2001:
LGTM. We can merge this.

Documenting why are we changing the CI:
- We should not be updating dependencies in CI. Hence we should remove the poetry update command
- To install poetry, following the official documentation, we should use the following command: https://python-poetry.org/docs/#installing-with-pipx
    ```bash
        pipx install poetry
    ```

jainmohit2001 (approved):

PR #695, opened 2024-12-10
Relax GetConfigurationResponse value to allow integer also

emlynmac:

Relaxes the validation schema on the GetConfigurationResponse to allow numerical values in addition to string values.

### Current behavior

Currently, only strings are allowed. Some devices are returning integers, which causes validation failure.

### New behavior

Integers are allowed to be a valid format for the value of the response.

### Impact

Enables use without modification for devices that pass integers directly, rather than encapsulated in strings.

### Checklist

1. [ ] Does your submission pass the existing tests?
2. [ ] Are there new tests that cover these additions/changes? 
3. [ ] Have you linted your code locally before submission?

### DIFF
--- a/ocpp/v16/schemas/GetConfigurationResponse.json
+++ b/ocpp/v16/schemas/GetConfigurationResponse.json
@@ -1,39 +1,36 @@
 {
-    "$schema": "http://json-schema.org/draft-04/schema#",
-    "title": "GetConfigurationResponse",
-    "type": "object",
-    "properties": {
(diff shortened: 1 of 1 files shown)


proelke:
Hi @emlynmac, these JSON schemas are the official ones from the OCA and we don't have any intention of modifying them. The changes you are asking for do not comply with the specification. I would reach out to the vendor to follow the spec properly or to use the skip validation parameters to not validate against the schema.

PR #759, opened 2026-06-02
Added Hz as a valid enum

Mforcen:
### Added Hz as valid units

I know that Hz is not a standard unit, but it some chargers spit measurements with those units, and since the library raises an error I assumed that this could be added, in the same way that Hertz was added to the schema, since it is not part of the standard.

Could we merge this?
Best regards.

### DIFF
--- a/ocpp/v16/schemas/MeterValues.json
+++ b/ocpp/v16/schemas/MeterValues.json
@@ -126,7 +126,8 @@
                                         "Celsius",
                                         "Fahrenheit",
                                         "Percent",
-                                        "Hertz"
+                                        "Hertz",
+                                        "Hz"
                                     ]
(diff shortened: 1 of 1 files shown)


mdwcrft:
I believe Hertz was added to the errata v4.0 (see it mentioned here https://github.com/mobilityhouse/ocpp/pull/70) so is valid but Hz has not been so wouldn't strictly be correct for OCPP 1.6

Mforcen:
I see, I did not read the errata, I assume that OCA would update their documentation, but maybe they keep the spec as they was released. Would it be possible to add a flag to skip or add non standard measurement values? I'm still willing to do work if needed.
Best regards, 
Manuel.

